PR #5, opened 2021-03-08
Format code with black and isort

meatballs:


s-cork:
We should include this as a pre-commit hook or using github actions.

PR #6, opened 2021-03-09
loading spinner color

s-cork:
This is a little idea for a function to change the loading spinner color to match your theme

To add it to a project put the following in native libraries

```html
<script src="https://cdn.jsdelivr.net/gh/anvilistas/anvil-extras/theme/assets/loading-spinner.js" color="#E91E63"></script>
```

The color transformation isn't perfect but it's reasonably. I thought it was cute...


meatballs:
So far, we've managed to avoid having anything in the them directory because that's not available from a dependency. Is there some way to implement this within a component instead, perhaps?

s-cork:
I started with a pure python implementation... but the spinner spins before your app loads. 😞 
So by the time you've imported the module - it's too late...

We wouldn't be adding anything to native libraries in this repo. The end user - would choose to include the CDN for this specific functionality. So it's never forced upon them and we keep our native libraries free.

The actual JS file can live anywhere in this repo. 
We just would need to document the cdn link somewhere.

meatballs:
right. got it! I'd misunderstood the cdn/native libraries stuff...

PR #14, opened 2021-03-11
Fix docs title for consistency

meatballs:


s-cork:
I think 1 liners like this can maybe just be self merged - with a comment like - small change merging...

PR #31, opened 2021-03-12
Switch Component

### LINKED ISSUE #12: Add a switch component
Probably taking inspiration from:

https://anvil.works/forum/t/another-switch-toggle-using-checkbox-and-anvil-roles/3832
https://anvil.works/forum/t/toggle-switch-demo-custom-html/755

meatballs:
closes #12


meatballs:
Mostly based on https://anvil.works/forum/t/another-switch-toggle-using-checkbox-and-anvil-roles/3832

s-cork:
This is an example of something that should be easy to do but isn't. 😞 
I should use this as an example of how to improve CustomComponents.

I've also been working on this haha. I've got some ideas.

First we can abuse the yaml and change the Container to checkbox.
It's crazy but it works... And also means you barely have to do anything with properties or events just include them for the design view.

`checked_colour` -> `checked_color`
We should be able to account for a `theme:color` in the `checked_color` attribute.
This is a bit tricky but anvil stores the themeColors in `window.anvilThemeColors` so once you know that you can do what anvil.

It would be nice if the `checked_color` was dynamic (I have an implementation of this)

The checkbox should be able to change size. In the forum example it would change size based on the font-size. 
The forum example also has a text property just like the checkbox does because really it's just a checkbox.

Happy to build on this if you like.

s-cork:
ooh and... we can just ignore the template in the Switch class and inherit straight from CheckBox...
Remove the middle man.

s-cork:
In terms of look and feel - what did you think of this version
https://materializecss.com/switches.html

meatballs:
Go for it. I like the ideas and that example is nice.

meatballs:
You still interested in doing this one or should I pick it up again?

s-cork:
yeah - it's still on my todo - i got a little frustrated with it being something that should just be a checkbox plus some css and not being able to make it as easy as when you have access to roles.

I'll make it my next pr.

s-cork:
Ok that's done I think. 
My only question is if you prefer the look of this one https://materializecss.com/switches.html
Or happy with this component as is.
Again worth checking it out in the designer 😉

meatballs:
I do prefer the the other look, but it's marginal. This is such a massive improvement over what's currently available that it hardly seems worth the bother.

meatballs:
I'm happy to merge this as is. If we want the other look later on, we can always open a new issue.

PR #34, opened 2021-03-16
Add margin_top property to PageBreak

### LINKED ISSUE #33: Fix margin on PageBreak component
The margin of -55px taken from the original css doesn't seem to work now that it's incorporated into a form.

Probably best to make it a property so it can be passed in.

meatballs:
closes #33


client_code/PageBreak/__init__.py
```diff
@@ -22,13 +22,35 @@
...
 
 class PageBreak(PageBreakTemplate):
-    def __init__(self, **properties):
+    def __init__(self, margin_top, **properties):
+        self.html = f"""
+<div class="break-container" style="overflow: hidden;">
+  <div style="page-break-after:always;"/>
+  <div style="margin-top: {margin_top};"/>
+</div>
+"""
```
s-cork, line 55: how's this as an alternative

```python
from anvil.js.window import jQuery as _S

def __init__(self, margin_top, **properties):
        self.margin_node = _S(anvil.js.get_dom_node(self)).find('.margin-element')
        # add the margin-element class to the relevant dom node

@property
def margin_top(self):
    return self._margin_top
@margin_top.setter
def margin_top(self, value):
    self.margin_node.css('margin-top', value)  
    self._margin_top = value

```

jQuery is useful here because it deals with numbers and turns them to pixels...

client_code/PageBreak/form_template.yaml
```diff
@@ -1,12 +1,9 @@
 container:
   type: HtmlTemplate
   properties: {tooltip: '', background: '', foreground: '', border: '', visible: true,
-    role: null, html: "<!--\nBased on the snippet at https://anvil.works/forum/t/plots-in-pdf-being-divided-between-two-pages/7774/5\n\
-      -->\n<div class=\"break-container\" style=\"overflow: hidden;\">\n  <div style=\"\
-      page-break-after:always;\"/>\n  <div style=\"margin-top: -55px;\"/>\n</div>\n\
-      \n<style>\n  .break-container {\n    border: 1px solid grey;\n  }\n  @media\
-      \ print {\n    .break-container {\n      border: none;\n    }\n  }\n</style>"}
+    role: null, html: ''}
```
s-cork, line 4: If you go with the above version I'd leave the html and css in - it makes it nicer in the design view at least.

client_code/PageBreak/__init__.py
meatballs: nice

meatballs:
Sorry, forgot to squash the damned "saved version" commit

s-cork:
think we've lost the css

meatballs:
I really ought to concentrate on one thing at a time!

meatballs:
This one is broken. I'll start another and do a proper job...

PR #39, opened 2021-03-25
Auto Refreshing Item

meatballs:
- [x] Implementation Code
- [x] Documentation
- [x] Example use in demo
- [x] override dict.update

### DIFF
--- a/client_code/Demo/__init__.py
+++ b/client_code/Demo/__init__.py
@@ -22,14 +22,17 @@
 # SOFTWARE.
 #
 # This software is published at https://github.com/anvilistas/anvil-extras
+from ..utils import auto_refreshing
 from ._anvil_designer import DemoTemplate
 
 __version__ = "1.0.0"
 
 
+@auto_refreshing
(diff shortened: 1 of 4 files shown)


s-cork:
Not necessarily part of this pr but we could make utils a lazy module and use the `__dir__` to expose the available attributes. I think meredydd added the feature so that a module with a `__dir__` comes up in the autocomplete

s-cork:
I don't know how important it is but overriding `__setitem__` isn't enough to trigger a refresh if you use `dict.update`. Might be a nice bonus.

meatballs:
> Not necessarily part of this pr but we could make utils a lazy module and use the `__dir__` to expose the available attributes. I think meredydd added the feature so that a module with a `__dir__` comes up in the autocomplete

I think this one ought to be a separate issue and pr.

s-cork (approved):
Lgtm

meatballs:
I'll put a new release together some time today.

PR #52, opened 2021-05-23
Implement Quill Editor

### LINKED ISSUE #51: Quill component
We could make this properly!

The quick start guide for anvil.js was nothing more than an intro.
Examples on the forum are mostly half baked

It'd be nice to have a Quill implementation that was more compatible with anvil.

```python
quill = Quill()
quill.getText()
quill.setText()
quill.getContent()
## these are all the js methods
## We'd just need to override the __getattr__
## we could regex for snake to camel case if we like


quill.to_simpleobject() # this is very easy and means you can save it in the database as a simple object
quill.to_markdown() # I'm sure this must be possible! Then you can just display it in a rich text component


## to thing about
quill.readonly
quill.toolbar_options

```

s-cork:
closes #51 

This is a wrapper around the quill editor.

Somethings I wasn't sure about
- is `content` the right name?
- Should I include the `get_text`, `set_text` functions just for the autocompleter?
  - currently they'll work when you call them but they won't show up. 
  - The docs might be sufficient here.

### TODO
- [x] add docs


s-cork:
You might wanna check this one out in the editor - I think i've managed to make the designer update with the property changes 😉

PR #56, opened 2021-05-25
Switch with materialize css

s-cork:
changes the css to use the materialize version

it doesn't change size in the same way as the other one - but that's fine I think.

I've added pre and post text.

colors dynamic in the same way
Screenshot 2021-05-25 at 23 03 55
Screenshot 2021-05-25 at 23 04 04


meatballs:
Nice! Could we just have an entry in the changelog and this looks good to go.

PR #59, opened 2021-05-30
Remove get_uid where possible

meatballs:
I've removed get_uid entirely from the determinate progress bar but I can't find a clean way to replace its use for the pseudo element on the indeterminate bar.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,11 +1,20 @@
 Unreleased Changes
 ------------------
+
+New Features
+============
 * `Update styling of switch component <https://github.com/anvilistas/anvil-extras/pull/56>`_
 * `Include pagination_click event in augment module <https://github.com/anvilistas/anvil-extras/pull/55>`_
+
+Changes
+=======
+* `Refactor of progress bars <https://github.com/anvilistas/anvil-extras/pull/59>`_
 
 v1.2.0 25-May-2021
 ------------------
 
+New Features
+============
 * `component.trigger('writeback') <https://github.com/anvilistas/anvil-extras/pull/47>`_
 * `MultiSelectDropDown component <https://github.com/anvilistas/anvil-extras/pull/44>`_
 * `@wait_for_writeback decorator <https://github.com/anvilistas/anvil-extras/pull/50>`_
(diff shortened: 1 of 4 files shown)


meatballs:
It might be worth refactoring that function to something like add_pseudo_element_style if that's its one remaining legitimate purpose.

s-cork:
```python
session.style_injector.inject(ProgressBar.css)
css = """
.anvil-role-indeterminate-progress-indicator:before {
  background-color: var(--track-color);
}
"""
session.style_injector.inject(css)

class Indeterminate(IndeterminateTemplate):
    def __init__(self, track_colour, indicator_colour, **properties):
        dom_node = get_dom_node(self)
        dom_node.style.setProperty("background-color", indicator_colour)
        dom_node.style.setProperty("--track-color", track_colour)

        self.role = "progress-track"
        self.indicator_panel.role = "indeterminate-progress-indicator"
        self.indicator_panel.background = indicator_colour
        
```

you can use the css var trick

meatballs:
Another option might be a function to set (and return) the id of the dom node.

meatballs:
The point of the uid is to ensure that the property is only set for the single instance of the component so it can cope with multiple independent instances on a single form. I think that code will set the same track colour for all instances.

s-cork:
It won't. css vars inherit from parents and aren't global.
https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#inheritance_of_custom_properties

I've double checked in the live demo with 2 indicators.

meatballs:
Nice! Hadn't realised that.

I think I still like the idea of a function to set a component's id so it can then be used in css directly without the css vars. You?

s-cork:
I think it's a useful function.

Thinking out loud - maybe get_uid can be adjusted to receive a component and return the id.
Then using the `functools.cache` to always return the same id if we've seen the component before...

I used the css var in Switch. But injecting individual css with a uid works too.
I prefer the var when the attribute can be set at runtime. 
It feels easier to set the custom css property dynamically than to adjust the css style sheet that we injected in the init.

meatballs:
Yep, I think both have their place. I think the id works better here as its only set once in the init but I agree that's ugly for anything set again later.

Nice to find some patterns!

I'll fix this up later...

PR #66, opened 2021-06-02
Implement Designer Components

### LINKED ISSUE #62: designer components
In #60, #31 and #51  i ended up writing a heap of javascript that was only for the designer.

One thing that i'd like to explore is moving this code into its own javascript files.
Either as a folder in this library or as a separate library within anvil-extras

Reasons:

**Size:**
- This code is for the designer only and isn't necessary when the app is being run.
- But the live app will still pull in the html as text unnecessarily 
- (the javascript never runs because it gets discarded before the component is added to the page)

**DRY**
- A lot of the javascript is the same so it would be worth pulling out the common routines to make future designer compatible components easier to write

**Growth**
- Development in the js eco system build process 😨 

**Editor**
- it's a pain working with anvil's html editor

The html would then look something like
```html
<div class="anvil-slider"></div>
<script type="module">
  import { DesignerSlider } from 'https://jspm.dev/anvil-extras/designer-components';
  DesignSlider.init();
</script>
```

This would keep the html small so reduce the code size. The javascript would only be loaded in the designer since in the real app we remove the script element before the component is added to the screen.

s-cork:
closes #62 

## TODO
- [x] - move out of theme.js and into a separate folder
- [x] - make it an npm package or host it on deno.land
- [x] - work out which cdn service to use to import the modules since we can't use `./_/theme...` when it's a dependency
          possibilities - skypack, jsdeliver.esm, jspm.den, deno.land/x
- [x] - use a js build tool esbuild/rollup 
- [x] - consider using a transpiler/minifier as part of the build


meatballs:
Could we just have an entry in the changelog for this and I reckon it's good to go?

PR #68, opened 2021-06-05
Implement Chip component

### LINKED ISSUE #19: Tags
I've written something similar before. A nice component for a lot of uis

s-cork:
uXbOl7gwRu

Inspired by https://materializecss.com/chips.html

Is tags an ok name - as in - it doesn't clash too much with `.tag` which is an anvil thing.
I could use `Chips` instead like the material design does?

Lots of possible extra features could be added:
- [x] primary_placeholder: string - when no tags are visible
- [x] secondary_placeholder: string - when 
- [x] ~include_input: boolean - This would be nice because then you could create your own input~
- [x] allow the tag to have an icon/image
- [x] ~allow a tag to be a list of dicts similar to `[{"text": "foo", "icon": "fa:bar", "close": False}, ...]`~
- [x] ~`tag_close`~ `close_icon` whether to include the close icon

methods:
- ~`add_tag`~
- ~`remove_tag`~

Any thoughts?

TODO:
- [x] Finalize api
- [x] docs

(I'm tempted to leave it as it is and wait for feature requests)

Closes #19


meatballs:
I do think the name clashes with tag. I'd go with chips instead.

And yes, let's get it in as is and wait to see what gets requested.

s-cork:
Yes agree - I think what i'll do is separate it into two components
`Chips` and `ChipsInput`.
That way the `Chips` component itself can be used in a variety of settings. And is basically just a styled Label with a couple of click events. The left icon can also be a circle when present to allow for an image if desired.

PR #85, opened 2021-06-17
Add docs for enterprise installation (modified version of old installation docs)

daviesian:
Third party dependencies are not supported in Enterprise Installations of Anvil. These docs are a fairly small change from the old Anvil Extras installation docs, detailing the clone/pull/force-push approach.


meatballs:
Thanks @daviesian - but this will also need an entry in `docs/guides/index.rst` in order for it to be visible. Could you add that to the PR?

daviesian:
Certainly!

PR #92, opened 2021-06-23
Fix messaging.py unsubscribe bug

hugetim:
Currently, unsubscribe method does not unsubscribe.


meatballs:
Nice catch! Could you add a line in the changelog for this?

hugetim:
Thanks! Sure, though not for a few hours.

hugetim:
Ok, I added a line to the changelog.

PR #96, opened 2021-06-24
Document auto-refresh reference-breaking behavior and revise related Demo code

### LINKED ISSUE #89: Auto-refresher removes reference relationship
In the auto_refreshing Demo code, self.item is initialized like this:
```python
self.item = self.default_item = dict(tally=100, counter=0,)
```
There are buttons to change item["tally"] and item["counter"], and the label components displaying these values are auto-refreshed. Then there is another button that resets the tally and counter to the default values via this line:
```python
self.item = self.default_item
```
This works as intended, but the trouble is that this actually shouldn't work--and it indeed doesn't work without the auto-refreshing enabled. The reason is that self.item was initialized to be equal to self.default_item by *reference*, so any changes to self.item should also change self.default_item. The fact that it doesn't when auto-refreshing is triggered by changes to self.item seems to be a subtle bug (or undocumented side effect) in the auto-refreshing functionality.

hugetim:
Intended to fix #89

### DIFF
--- a/client_code/Demo/__init__.py
+++ b/client_code/Demo/__init__.py
@@ -25,7 +25,7 @@ class Demo(DemoTemplate):
             chips=["a", "b", "c"],
             text="",
         )
-        self.item = self.default_item.copy()
+        self.item = self.default_item
         self.init_components(**properties)
 
     def timer_1_tick(self, **event_args):
@@ -44,7 +44,7 @@ class Demo(DemoTemplate):
         self.item["counter"] += 1
 
     def reset_button_click(self, **event_args):
-        self.item = self.default_item.copy()
+        self.item = self.default_item
 
     ###### MULTI SELECT ######
     def multi_select_drop_down_1_change(self, **event_args):
--- a/client_code/Demo/__init__.py
+++ b/client_code/Demo/__init__.py
@@ -17,7 +17,7 @@ class Demo(DemoTemplate):
         self.init_custom_slider_formatter()
 
         self.progress = 0
-        self.item = self.default_item = dict(
+        self.default_item = dict(
             tally=100,
             counter=0,
             values=self.slider.start,
@@ -25,6 +25,7 @@ class Demo(DemoTemplate):
             chips=["a", "b", "c"],
             text="",
         )
+        self.item = self.default_item.copy()
         self.init_components(**properties)
 
     def timer_1_tick(self, **event_args):
@@ -43,7 +44,7 @@ class Demo(DemoTemplate):
         self.item["counter"] += 1
 
     def reset_button_click(self, **event_args):
-        self.item = self.default_item
+        self.item = self.default_item.copy()
 
     ###### MULTI SELECT ######
     def multi_select_drop_down_1_change(self, **event_args):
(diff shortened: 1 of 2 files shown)


client_code/Demo/__init__.py
```diff
@@ -17,14 +17,15 @@ def __init__(self, **properties):
...
-        self.item = self.default_item = dict(
+        self.default_item = dict(
             tally=100,
             counter=0,
             values=self.slider.start,
             agree=self.slider_agree.value,
             chips=["a", "b", "c"],
             text="",
         )
+        self.item = self.default_item.copy()
```
meatballs, line 28: Actually, I think the `copy()` is redundant here. Setting `self.item` will pass `self.default_item` to the `dict` function anyway.

client_code/Demo/__init__.py
```diff
@@ -43,7 +44,7 @@ def plus_button_click(self, **event_args):
         self.item["counter"] += 1
 
     def reset_button_click(self, **event_args):
-        self.item = self.default_item
+        self.item = self.default_item.copy()
```
meatballs, line 47: Again, I think `copy()` is redundant

client_code/Demo/__init__.py
meatballs: I know we discussed using `copy()` on #89  but I actually think it's just the docs that need updating.

docs/guides/modules/utils.rst
```diff
@@ -71,7 +71,19 @@ To use it, import the decorator and apply it to the class for a form:
        def __init__(self, **properties):
            self.init_components(**properties)
 
-Now, the form has an ``item`` property which behaves like a dictionary. Whenever a value of that dictionary changes, the form's ``refresh_data_bindings`` method will be called.
+Now, the form has an ``item`` property which behaves like a dictionary. Whenever a value of that dictionary changes, the form's ``refresh_data_bindings`` method will be called. 
+
+Note: The ``item`` property will no longer reference the same object after an auto-refresh. Rather, in the following example, it is as though auto-refresh adds the ``item = dict(item)`` line:
```
meatballs, line 76: Can we remove `after and auto-refresh` from this sentence? This happens when the form is instantiated rather than when a refresh occurs.

meatballs (changes requested):

client_code/Demo/__init__.py
hugetim: It is redundant with auto_refreshing. But in my original demo app, I was going to show the example with and without auto-refreshing to demonstrate the difference. If others try the same, they may be confused by the lack of reset behavior without auto-refreshing. On the other hand, the exercise will highlight to them the now-documented auto-refreshing side-effect that causes `self.item` to no longer reference the same underlying object as `self.default_item`.

If, hearing my reasoning for wanting to include `copy()`, you'd still rather not, I'll make the change to remove them.

meatballs: I think I'd prefer to leave it out. In your demo app, it might be worth just chucking in a comment or two to explain the difference (perhaps with a link to the docs).

meatballs (approved):

PR #99, opened 2021-06-26
Add navigation.emulate_link_click

### LINKED ISSUE #88: Navigation click_link(target="") method?
I would like to be able to emulate clicking a navigation menu link from code. The undocumented method 'open_form' gets me most of the way there but it doesn't show the menu link as selected.

hugetim:
Closes #88

### DIFF
--- a/client_code/navigation.py
+++ b/client_code/navigation.py
@@ -39,7 +39,7 @@ def get_form(name, *args, **kwargs):
     try:
         return _forms[name]["class"](*args, **kwargs)
     except KeyError:
-        raise KeyError("No form registered under that name")
+        raise KeyError("No form registered under name: " + str(name))
 
 
 def open_form(form_name, full_width=False):
@@ -56,7 +56,7 @@ def go_to(target):
         if link.tag.target == target:
             break
     else:  # no break
-        raise KeyError("No menu link matching that target")
+        raise ValueError("No menu link matching target: " + str(target))
     link.raise_event("click")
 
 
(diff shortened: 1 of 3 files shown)


hugetim:
Looks like I need to bite the bullet and get black installed. There's also probably a better name for this function.

s-cork:
This might work for you. 

```
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

client_code/navigation.py
```diff
@@ -46,6 +46,12 @@ def open_form(form_name, full_width=False):
     get_open_form().content_panel.clear()
     get_open_form().content_panel.add_component(form, full_width_row=full_width)
 
+    
+def emulate_link_click(target):
+    """Emulate clicking a menu link"""
+    matching_links = [link for link in _links if link.tag.target == target]
+    matching_links[0].raise_event('click')
+
```
s-cork, line 54: A nicer error when there are no matching links. 

You can also do a loop and break out when you find the first one. And use the no break syntax. 

```python
for link in links:
    if condition:
        break
else: # no break
    raise ...
link.raise_event('click')
```

Re name of the function. `go_to`or `navigate_to`?

s-cork (approved):
@meatballs  - LGTM - over to you.

client_code/navigation.py
hugetim: I implemented this, changed name to go_to, and also tried adding a more helpful error message to the get_form function.

hugetim:
Ok, I'm stopping with the error message fiddling now.

meatballs:
Could we have some docs and a changelog entry?

PR #108, opened 2021-07-08
Update __init__.py

juanpaul101:


s-cork:
@juanpaul101  if you don't want the border radius you can change your theme.css

```css
.anvil-role-autocomplete {
    border-radius: 0px !important;
}
```

PR #112, opened 2021-08-23
bump no ui slider version

### LINKED ISSUE #84: update noUiSlider
They just release a new version which adds an api that I contributed with anvil in mind 😉

s-cork:
doesn't change any behaviour but does make getting the raw values easier since we can now use noUiSlider's public api to do so.
https://github.com/leongersen/noUiSlider/pull/1145

the bundle changes are just from running the build script for the javascript.

close #84


meatballs:
LGTM. Could we just have a changelog entry?

s-cork:
Sure. I'll do that tomorrow and merge.

PR #117, opened 2021-08-28
Support quill modules

### LINKED ISSUE #87: Quill Image Resize Module
Hi! Amazing and very useful tools so congrats and thanks for your work guys!!

One question.. How would I go about adding the Quill Resize module when using the extras? I managed to get it working when adding Quill by myself by adding the line below it to native libraries and making the necessary changes to the JS setup.

I tried adding the line to native libraries but it did nothing sadly as I guess I'd have to change the Quill setup on JS.

Native libraries line: `<script src="https://cdn.jsdelivr.net/npm/quill-image-resize-module@3.0.0/image-resize.min.js" async></script>`
Link to Quill image resize module: https://github.com/kensnyder/quill-image-resize-module

I don't know anything other than Python so sorry if this is easy, I just haven't found the way yet. Thanks!

s-cork:
close #87 

The code to implement #87 might include Quill and the imageResize module scripts in Native Libraries.
(It would need to include both since imageResize relies upon Quill existing)

Then in a form that uses a Quill component.

```python
class Form1(Form1Template):
    # Set Form properties and Data Bindings.
    self.quill.modules = {"imageResize": {}}
```

@sebasm611 it's a bit late but how's this api for you?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## New Features
+* Quill - dynamically add custom modules
+  https://github.com/anvilistas/anvil-extras/pull/117
 ## Bug Fixes
 * Autocomplete - can now be used inside an alert
   https://github.com/anvilistas/anvil-extras/pull/114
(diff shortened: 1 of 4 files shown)


sebasm611:
Perfect! Thank you.. this will be very useful

PR #129, opened 2021-09-07
Address storage access being denied

### LINKED ISSUE #106: Storage error (only) when running in development mode
When clicking the "Run" button to test an Anvil app in development mode (but not when running the app via its url), I get the following error when trying to import `local_storage`:
```
ExternalError: SecurityError: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
at app/anvil_extras/storage.py, line 103
  called from StorageDemo, line 5
  called from Main, line 9
```
I suppose it's something to do with the iframe? This was not happening just 24 hours ago, but now it persistently is.

Clone link: https://anvil.works/build#clone:X32NITRHDRRHDOFO=OIDA3G3GQVMLMMT2LPC7BV3O

hugetim:
Resolves #106 

I'm unsure if I am catching the error correctly. Note: I added a comment there to prevent flake8 from tripping on 'ExternalError'.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
   https://github.com/anvilistas/anvil-extras/pull/116
 * Popover - documentation added and clone link updated. The example now imports `anvil_extras`
   https://github.com/anvilistas/anvil-extras/pull/121
-* storage - use dictionary alternative when storage is not available, printing a warning
+* storage - use dictionary alternative when storage not available in Anvil's 'debug' mode, printing a warning
   https://github.com/anvilistas/anvil-extras/pull/129
 
 # v1.5.2 23-Aug-2021
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
   https://github.com/anvilistas/anvil-extras/pull/116
 * Popover - documentation added and clone link updated. The example now imports `anvil_extras`
   https://github.com/anvilistas/anvil-extras/pull/121
+* storage - use dictionary alternative when storage is not available, printing a warning
+  https://github.com/anvilistas/anvil-extras/pull/129
 
 # v1.5.2 23-Aug-2021
 
--- a/client_code/storage.py
+++ b/client_code/storage.py
@@ -109,8 +109,16 @@ class Storage:
             self[key] = value
 
 
-local_storage = Storage(_window.get("localStorage"))
-session_storage = Storage(_window.get("sessionStorage"))
+try:
+    local_storage = Storage(_window.get("localStorage"))
+    session_storage = Storage(_window.get("sessionStorage"))
+except ExternalError:  # noqa
+    print(
+        "Warning: Access to storage denied, likely due to browser permissions."
+        + " Using dictionary alternative which will not persist between browser sessions."
+    )
+    local_storage = {}
+    session_storage = {}
 
 
 if __name__ == "__main__":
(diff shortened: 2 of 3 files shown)


meatballs:
I'm not sure about this. I'm not keen on suppressing an error and carrying on with only a print statement. There may well be situations we haven't thought of that raise that error in production systems where the print would not even be seen.

I'd far prefer to test for known conditions and raise custom errors if those are found (e.g. Running within the ide).

hugetim:
Yeah, I agree. I was feeling pretty uncomfortable with this as I thought more about how it could play out.

Shall I take a crack at Stu's proposed [option 2](https://github.com/anvilistas/anvil-extras/issues/106#issuecomment-873729342), then?

meatballs:
How about...

Keep the try except as you have it.

In the except clause, test for whether the app is being run in the ide.

If so, do as per your current solution.

If not, raise a custom error (which will need to be defined).

I also don't like the noqa. If the linter is complaining, we should work out why and fix it. Is that error class imported?

(sorry if this pseudo code is unclear. I'm on my phone).

stenci:
Considering that:
- Any change done at this point, including the one just committed, will not be backward compatible
- Some apps will be happy to deal with non persistent `local_storage` and some will not, so it's impossible to make everybody happy
- I (and the Zen of Python) like explicit better than explicit

I would create two different modules with different behaviors. Then the developer will be able to pick the best.

I would keep the exception so there is no broken backward compatibility, but I would change it to an understandable one. 
Then I would add a new module with a new global object with the new behavior.
Then the developer will decide which one to use. They can pick one of these:
```
from anvil_extras.storage import local_storage
from anvil_extras.storage_or_dict import local_storage
```

The `storage` module will raise an understandable exception.
The `storage_or_dict` module will fall back to the dict.
Both the modules will import the same third module with the definition of the `Storage` class.

hugetim:
I like that idea and I think it's clear.

The only thing I'm stuck on is the ExternalError. I don't know where it's from. I assume it's something Anvil-specific, but I can't find it documented anywhere (or when I mess around with autocomplete). All I really know is that this shows up in the app logs when trying to `import storage`: 
```
ExternalError: SecurityError: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
at app/anvil_extras/storage.py, line 101
```

stenci:
I got a few `ExternalError`'s while playing with popovers. I was creating and deleting popovers too quickly, some javascript function (bootstrap?) was trying to do something with them and crashing because they didn't exist anymore. And Anvil was reporting them as `ExternalError`.

In this case, again javascript complains about something going wrong, Anvil doesn't know what to do with that complaint and raises an `ExternalError`.

My guess is that an `ExternalError` is the browser equivalent of a server's `InternalError`.

If the current `ExternalError` is clear enough, then no need to change it. Otherwise it could be caught just to raise a nicer one.

hugetim:
Well, `ExternalError` doesn't actually work, as it turns out. I hadn't actually tested that code. Now, when I try to run something similar, I get: `NameError: name 'ExternalError' is not defined`

s-cork:
I would raise a `RuntimeError` if I had to chose an error.

`ExternalError` happens anytime an error was caused from javascript rather than python.
But it's a python object whose base class is `Exception` so you can still catch it here.
https://github.com/anvilistas/anvil-extras/issues/106#issuecomment-915360390

Issue #40, opened 2021-04-02
Add docs for hashrouting


s-cork:
Need to do this. Adding it here as a reminder!

hugetim:
I guess the main challenge is to translate the docs from markdown [here](https://github.com/s-cork/HashRouting/blob/master/README.md) to Restructured Text [here](https://github.com/anvilistas/anvil-extras/tree/main/docs/guides/modules).

s-cork:
yeah that's definitely one of the factors holding this back! 
Plus I think it could do with some tweaks and it's a lot of text!

Also @meatballs had some thoughts about how we might structure our docs. 
So part of me was thinking that the routing module docs might be a good place adopt the style.
But maybe it's best just to get them in and tidy them later.

PR #132, opened 2021-09-10
include routing.alert api

### LINKED ISSUE #131: Routing.alert
See forum post. 

https://anvil.works/forum/t/how-to-close-alerts-when-navigating-away-from-current-page-with-hashrouting/9711/

Correct behaviour should probably be to close alert if the alert is dismissible and to prevent unloading if it's not dismissible.

s-cork:
close #131 

wrapper around anvil alerts.
If an alert is dismissible then the alert will be closed on navigation
if an alert is not dismissible then navigation will be halted

---

Also included in this pr - logger doesn't save logs by default - i figure this feature is not really used and seems unnecessary to create an ever growing list of logs.

---

I'll document this as part #40 

@stenci might want to take a look at the api/code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## New Features
 * Quill - dynamically add custom modules
   https://github.com/anvilistas/anvil-extras/pull/117
+* routing - routing.alert wraps anvil.alert to ensure dismissible alerts are closed on navigation
+  https://github.com/anvilistas/anvil-extras/pull/132
 ## Bug Fixes
 * Autocomplete - can now be used inside an alert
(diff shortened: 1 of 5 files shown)


stenci:
This works well with confirm and non dismissible alert, but doesn't close the dismissible alert when it navigates away.

I tested it with this code:
```python
  def title_click(self, **event_args):
    print(confirm('confirm'))
    print(alert('non dismissible alert', dismissible=False))
    print(alert('dismissible alert'))
```

It does prevents the navigation on the first two rows (good), it allows the navigation on the third (good), but doesn't close the alert (bad).

s-cork:
oops - forgot a line of code. That should now be fixed

stenci:
It works well now

PR #135, opened 2021-09-10
Use localforage to create a wrapper around indexed_db

### LINKED ISSUE #106: Storage error (only) when running in development mode

### LINKED ISSUE #130: Storage - IndexedDB storage
localStorage is good for some things but it's fairly limited.
We might consider a wrapper for IndexedDB.

There are plenty of good javascript wrappers already out there.
https://github.com/localForage/localForage
or 
https://github.com/dannyconnell/localbase

So we could wrap a wrapper?

s-cork:
close #130, close #106, close #129 

I use `localforage` to wrap `localStorage` and `indexedDB`.

This pr deprecates `session_storage` - it's not particularly useful anyway.
It's deprecated because localforage doesn't have a wrapper for it.

It adds a nice api for creating a new store
This is like creating a new dictionary object.
```python
from anvi_extras.storage import indexed_db as db
todo_store = db.get_store('todos')
```

It also adds an `is_supported()` method which returns a `boolean`.
You don't get an error on import anymore.
But if you try to use the store in anyway, and it's not supported you do get 
`ExternalError: Error: No available storage method found.`

compromises and hacks:
- the repr can't expose the internals since localforage is very asyncronous and reprs can't be asyncronous in skulpt
- I had to hack the return values to convert proxy objects back to dictionaries
- I used a Promise hack for removing an item because of asyncronous code (del doesn't like being asyncronous either)

### DIFF
--- a/client_code/storage.py
+++ b/client_code/storage.py
@@ -17,20 +17,21 @@ _Proxy = type(_window)
 _Object = _window.Object
 
 
-def to_py(obj):
-    """converts simple proxy objects (and nested simple proxy objects) as dictionaries"""
+def _deserialize(obj):
+    """convert simple proxy objects (and nested simple proxy objects) to dictionaries"""
+    # @TODO datetime objects - we'd also need a _serialize method for that
     if isinstance(obj, list):
         ret = []
         for item in obj:
-            ret.append(to_py(item))
+            ret.append(_deserialize(item))
         return ret
     elif type(obj) is _Proxy and obj.__class__ == _Object:
         # Then we're a simple proxy object
-        # keys are strings so only to_py the values
+        # keys are strings so only _deserialize the values
         ret = {}
         # use _Object.keys to avoid possible name conflict
         for key in _Object.keys(obj):
-            ret[key] = to_py(obj[key])
+            ret[key] = _deserialize(obj[key])
         return ret
     else:
         # we're either bytes, str, ints, floats, None, bool
@@ -58,7 +59,7 @@ class StorageWrapper:
 
     def __getitem__(self, key):
         if key in self._store.keys():
-            return to_py(self._store.getItem(key))
+            return _deserialize(self._store.getItem(key))
         raise KeyError(key)
 
     def __setitem__(self, key, val):
@@ -88,11 +89,13 @@ class StorageWrapper:
 
     def items(self):
         """returns the items for local storage as an iterator"""
-        return ((key, to_py(self._store.getItem(key))) for key in self._store.keys())
+        return (
+            (key, _deserialize(self._store.getItem(key))) for key in self._store.keys()
+        )
 
     def values(self):
         """returns the values for local storage as an iterator"""
-        return (to_py(self._store.getItem(key)) for key in self._store.keys())
+        return (_deserialize(self._store.getItem(key)) for key in self._store.keys())
 
     def put(self, key, value):
         """put a key value pair into local storage"""
--- a/client_code/storage.py
+++ b/client_code/storage.py
@@ -39,9 +39,10 @@ def _deserialize(obj):
 
 
 class StorageWrapper:
-    def __init__(self, store, driver, name):
+    _driver = None
+
+    def __init__(self, store, name):
         self._store = store
-        self._driver = driver
         self._name = name
 
     def is_available(self):
@@ -75,7 +76,7 @@ class StorageWrapper:
 
     def __repr__(self):
         # we can't print the items like a dictionary since we get a SuspensionError here
-        return f"<{driver_names[self._driver]} for {self._name!r} store>"
+        return f"<{self.__class__.__name__} for {self._name!r} store>"
 
     def __iter__(self):
         return self.keys()
@@ -132,18 +133,22 @@ class StorageWrapper:
             return NotImplemented
         return self._driver == other._driver and self._name == other._name
 
-    def get_store(self, name):
+    @classmethod
+    def create_store(cls, store_name: str):
         """
-        Create a new storage object. Inside either local_storage or indexed_db.
-        e.g. todo_store = indexed_db.get_store('todos')
-        message_store = indexed_db.get_store('messages')
+        Create a new storage object. Inside the browser's IndexedDB or localStorage.
+        e.g. todo_store = indexed_db.create_store('todos')
+        message_store = indexed_db.create_store('messages')
         """
-        return StorageWrapper(
-            self._store.createInstance(
-                {"storeName": name, "driver": self._driver, "name": "anvil_extras"}
+        return cls(
+            _forage.createInstance(
+                {
+                    "storeName": store_name,
+                    "driver": [cls._driver, "failDriver"],
+                    "name": "anvil_extras",
+                }
             ),
-            self._driver,
-            name,
+            store_name,
         )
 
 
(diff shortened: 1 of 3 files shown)


docs/guides/modules/storage.rst
```diff
@@ -49,78 +53,88 @@ Change the theme at startup
...
+   both :attr:`indexed_db` and :attr:`local_storage` are instances of a dictionary like :class:`StorageWrapper` class.
+
+   .. describe:: is_supported()
+
+      Check if the storage object is supported. Returns a ``boolean``.
+
+   .. describe:: get_store(name)
 
-   local_storage is a dictionary like object.
+      Get or create a ``storage`` object. e.g. ``todo_store = indexed_db.get_store('todos')``. This will create a new storeage object inside the browser's ``IndexedDB``.
```
hugetim, line 66: typo in line 66 of storage.rst: "storeage" -> "storage"

docs/guides/modules/storage.rst
```diff
@@ -49,78 +53,88 @@ Change the theme at startup
...
-      Return the item of :attr:`local_storage` with key *key*.  Raises a :exc:`KeyError` if *key* is
-      not in :attr:`local_storage`. Raises a :exc:`TypeError` if *key* is not a string.
+      Return the item of *store* with key *key*.  Raises a :exc:`KeyError` if *key* is
+      not in *store*. Raises a :exc:`TypeError` if *key* is not a string.
 
-   .. describe:: local_storage[key] = value
+   .. describe:: store[key] = value
 
-      Set ``local_storage[key]`` to *value*. The *value* must be a json compatible object.
+      Set ``store[key]`` to *value*. If the value is not a JSONable data type it may be stored correctly. e.g. a ``datetime`` object.
```
hugetim, line 84: line 84 of storage.rst: should "correctly" read "incorrectly" here?

s-cork: yeap

hugetim (approved):
This looks good to me (aside from the few docs typos noted), as far as I understand it. 

I like that the RuntimeError("browser storage is not available") doesn't get triggered until the store is accessed (as opposed to the current situation, in which it is triggered upon importing the storage module). This (assuming it indeed works this way) allows the app developer to catch that error as desired, in a more appropriate place in their code.

And it seems better not to catch the error, even in 'debug' mode.

s-cork:
@hugetim - thanks for the review!

I was hoping that `is_supported` would just work - but it doesn't help with #129 since `localforage` just checks for browser support generally and not security problems.

at the moment - the first time that a storage object is accessed you'll get
`ExternalError: Error: No available storage method found.`.

But there won't be any errors from importing the module anymore... 👍 

So there's a few options for this.
1. I can try to make `is_supported()` do a bit more and check the security. Probably just by calling any method from the `_store`
2. I can wrap all store methods inside a try catch block
3. both of the above
4. do nothing the api is fine like this.

I'd vote for 1.

hugetim:
I agree: 1 would be good.

meatballs:
I'm travelling for the next day or two so can't look at this properly right now.

All looks good from a cursory glance!

s-cork:
@hugetim I think I managed to do 2 without wrapping everything in `try/catch` see b1321d4. 
@meatballs no rush.

s-cork:
@meatballs - when you get to reviewing - one thing to to consider is that the `local_storage` prefix has changed. Which means that previous stored objects won't be accessible. I don't see this as a problem since the code that works with `local_storage` should cover the case when the key doesn't exist.

But if you think that'll be an issue then I can write a loop to move anything from local_storage into the new new world that comes with this pr. Something like...

```python

# temporary for backwards compatibility - remove in the next next release
try:
    old_prefix = "anvil_storage_"
    ls = _window.get("localStorage", {})
    for key in ls:
        if key.startswith(old_prefix):
            val = ls.getItem(key)
            ls.removeItem(key)
            local_storage[key] = val # move to the new local_storage object
except Exception:
    # probably a security error
    pass
```

Like I said - I don't think this is really a big issue given the nature of the `local_storage`.

meatballs:
Very nice piece of work! Could you just fix up the merge conflict and it's good to go.

s-cork:
great that's done - can you just check the api is ok - naming things.

```python
if not indexed_db.is_available():
    alert('ooops')

# and

todo_store = indexed_db.get_store('todos')
# gets the todo store or creates a new store if it doesn't exist
```

client_code/storage.py
```diff
@@ -5,79 +5,97 @@
...
+        return (
+            (key, _deserialize(self._store.getItem(key))) for key in self._store.keys()
+        )
 
     def values(self):
         """returns the values for local storage as an iterator"""
-        return self._map_store(lambda key: self.__getitem__(key))
+        return (_deserialize(self._store.getItem(key)) for key in self._store.keys())
 
     def put(self, key, value):
```
meatballs, line 100: using 'put' slightly implies a stack structure. How about 'store' or 'save'?

client_code/storage.py
```diff
@@ -99,43 +117,129 @@ def pop(self, key, default=None):
...
 
-
-local_storage = Storage(_window.get("localStorage"))
-session_storage = Storage(_window.get("sessionStorage"))
+    def __eq__(self, other):
+        if not isinstance(other, StorageWrapper):
+            return NotImplemented
+        return self._driver == other._driver and self._name == other._name
+
+    def get_store(self, name):
```
meatballs, line 133: This kind of factory function might be better as either a class method or a module level function - named something like 'create_store'

client_code/storage.py
rhurlbatt: FWIW - from a noobs perspective I also think that 'store' is less daunting.

s-cork: good idea - `store` it shall be. 
I'll alias `put` with `store` and remove `put` from the api docs.

client_code/storage.py
s-cork: Yeah - i think i'd prefer it to be a `classmethod` too. 
I just struggled to think of the best way to do that since I needed to know if you wanted an `indexed_db` store or a `local_storage` store.

How about - I subclass the `StorageWrapper` into `LocalStorageWrapper` and `IndexedDBStorageWrapper`

Yes That'd work. And it also makes things a bit clearer. `indexed_db` is just the default store instance of the `IndexedDBStorageWrapper`.

And @meatballs votes for `create_store`. Since that was your instinct for this method lets go with that.

meatballs: The other alternative is to use a module level factory function with the 'storage_type' as an argument. You then don't need the subclasses.

meatballs: (for me, inheritance is one of those situations where you sit on your hands for a moment and ponder whether there might be another way)!

s-cork: I can see that -  what would you vote the type of the `storage_type` to be?

meatballs: I'd make that a simple string parameter and either map the string to the relevant class within the function or use `getattr` and the relevant module to derive the class.

meatballs: I'd make that a string and map it to either `_indexed_db_wrapper` or `_local_storage_wrapper` within the function's body.

meatballs: or, I suppose, we could choose one of them to be the default (say, indexed_db) and use a boolean like `use_local_storage` with a default of `False`

s-cork: I'll sleep on it - I already wrote the inheritance version before I read:
> inheritance is one of those situations where you sit on your hands for a moment and ponder whether there might be another way

So i've pushed that for now: 0aa8577151d204542378cccddd54173b5a85ba12
Maybe take a look at it if you get a moment. See how much it makes you wince...

meatballs: It's absolutely fine. I'd have probably gone with a module level function, but the difference is marginal.

You done with this now? It looks ready to merge to me.

s-cork: I did a little clean up yesterday and played around with the module level function. (not yet commited)

Let's call it approved - and then i'll hit the merge button later today. (when i need a break from javascript)

meatballs (approved):

docs/guides/modules/storage.rst
```diff
@@ -45,82 +53,125 @@ Change the theme at startup
...
+        def save_todo_btn_click(self, **event_args):
+            if not self.todo_input.text:
+                return
+            id = str(uuid4())
+            todo = {"id": id, "todo": self.todo_input.text, "completed": False}
+            todo_store[id] = todo
+            self.todo_panel.items = self.todo_panel.items + [todo]
+            self.todo_input.text = ""
+
+
```
s-cork, line 84: @meatballs - can you just check through this code snippet - see if it makes sense - and whether it provides the right api.
I think it probably does.
There's a little bit of tedium in that you have to store the id as the key and in the value. But I think that's ok...

meatballs: LGTM

PR #139, opened 2021-09-14
Add border property to PageBreak

### LINKED ISSUE #142: HTML issues
turns out this bug is ours:
https://anvil.works/forum/t/pagebreak-component-not-working-after-anvil-update/9754/3

It's todo with our HTML being invalid and an Anvil update.
Basically self closing tags are invalid but they were previously parsed.

todo:
- [ ] go through our html and check that the html is valid

(When I fix the html for the pagebreak it works again)

stenci:
Sometimes the PDF contains horizontal grey lines, so the PageBreak lines get confused with the ones that are part of the PDF.

This new `border` property allows to change the default style to something different form the rest of the PDF. It has no effect on the generated PDF.

In one of my PDFs I'm using:
```python
self.add_component(TruckPDF(self.truck))
for crate in self.truck.crates:
  self.add_component(PageBreak(border='1px dashed red'))
  self.add_component(CratePDF(crate))
```

close #142

### DIFF
--- a/client_code/PageBreak/__init__.py
+++ b/client_code/PageBreak/__init__.py
@@ -15,8 +15,11 @@ __version__ = "1.5.2"
 class PageBreak(PageBreakTemplate):
     def __init__(self, margin_top=0, border="1px solid grey", **properties):
         self.margin_node = _S(anvil.js.get_dom_node(self)).find(".margin-element")
+        self.break_container = _S(anvil.js.get_dom_node(self)).find(".break-container")
+
         self.margin_top = margin_top
         self.border = border
+
         self.init_components(**properties)
 
     @property
@@ -34,5 +37,5 @@ class PageBreak(PageBreakTemplate):
 
     @border.setter
     def border(self, value):
-        self.margin_node.css("border", value)
+        self.break_container.css("border", value)
         self._border = value
--- a/client_code/PageBreak/__init__.py
+++ b/client_code/PageBreak/__init__.py
@@ -13,9 +13,10 @@ __version__ = "1.5.2"
 
 
 class PageBreak(PageBreakTemplate):
-    def __init__(self, margin_top=0, **properties):
+    def __init__(self, margin_top=0, border="1px solid grey", **properties):
         self.margin_node = _S(anvil.js.get_dom_node(self)).find(".margin-element")
         self.margin_top = margin_top
+        self.border = border
         self.init_components(**properties)
 
     @property
@@ -26,3 +27,12 @@ class PageBreak(PageBreakTemplate):
     def margin_top(self, value):
         self.margin_node.css("margin-top", value)
         self._margin_top = value
+
+    @property
+    def border(self):
+        return self._border
+
+    @border.setter
+    def border(self, value):
+        self.margin_node.css("border", value)
+        self._border = value
--- a/client_code/PageBreak/__init__.py
+++ b/client_code/PageBreak/__init__.py
@@ -14,8 +14,9 @@ __version__ = "1.5.2"
 
 class PageBreak(PageBreakTemplate):
     def __init__(self, margin_top=0, border="1px solid grey", **properties):
-        self.margin_node = _S(anvil.js.get_dom_node(self)).find(".margin-element")
-        self.break_container = _S(anvil.js.get_dom_node(self)).find(".break-container")
+        dom_node = _S(anvil.js.get_dom_node(self))
+        self.margin_node = dom_node.find(".margin-element")
+        self.break_container = dom_node.find(".break-container")
 
         self.margin_top = margin_top
         self.border = border
--- a/client_code/PageBreak/form_template.yaml
+++ b/client_code/PageBreak/form_template.yaml
@@ -3,10 +3,10 @@ container:
   properties: {tooltip: '', background: '', foreground: '', border: '', visible: true,
     role: null, html: "<!--\nBased on the snippet at https://anvil.works/forum/t/plots-in-pdf-being-divided-between-two-pages/7774/5\n\
       -->\n<div class=\"break-container\" style=\"overflow: hidden;\">\n  <div style=\"\
-      page-break-after:always;\"/>\n  <div class=\"margin-element\" style=\"margin-top:\
-      \ 0px;\"/>\n</div>\n\n<style>\n  .break-container {\n    border: 1px solid grey;\n\
-      \  }\n  @media print {\n    .break-container {\n      border: none !important;\n\
-      \    }\n  }\n</style>"}
+      page-break-after:always;\"></div>\n  <div class=\"margin-element\" style=\"\
+      margin-top: 0px;\"></div>\n</div>\n\n<style>\n  .break-container {\n    border:\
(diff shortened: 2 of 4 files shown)


client_code/PageBreak/form_template.yaml
```diff
@@ -4,11 +4,11 @@ container:
     role: null, html: "<!--\nBased on the snippet at https://anvil.works/forum/t/plots-in-pdf-being-divided-between-two-pages/7774/5\n\
       -->\n<div class=\"break-container\" style=\"overflow: hidden;\">\n  <div style=\"\
       page-break-after:always;\"/>\n  <div class=\"margin-element\" style=\"margin-top:\
-      \ 0px;\"/>\n</div>\n\n<style>\n  .break-container {\n    border: 1px solid grey;\n\
```
s-cork, line 7: Have you checked that there is still a border if you add a page beak in the design view?

It's difficult to read the yaml but it looks like this change has removed the design view border.

stenci: No, I had not checked (because I had never added a `PageBreak` from the IDE). Sorry.

This version works from both IDE and code. In the IDE it does not show the effect of the `border` property. I don't think Anvil allows to do that (... [unless you write some javascript?](https://anvil.works/forum/t/custom-component-in-design-mode/1194))

Is it possible for me to change the code in this PR?
Or do I need to create a new PR?
Or do you want the content of the files, and you change them on your side?

Here is the content of the two updated files:

**\_\_init\_\_.py**
```
# SPDX-License-Identifier: MIT
#
# Copyright (c) 2021 The Anvil Extras project team members listed at
# https://github.com/anvilistas/anvil-extras/graphs/contributors
#
# This software is published at https://github.com/anvilistas/anvil-extras
import anvil
from anvil.js.window import jQuery as _S

from ._anvil_designer import PageBreakTemplate

__version__ = "1.5.2"


class PageBreak(PageBreakTemplate):
    def __init__(self, margin_top=0, border="1px solid grey", **properties):
        self.margin_node = _S(anvil.js.get_dom_node(self)).find(".margin-element")
        self.break_container = _S(anvil.js.get_dom_node(self)).find(".break-container")

        self.margin_top = margin_top
        self.border = border

    @property
    def margin_top(self):
        return self._margin_top

    @margin_top.setter
    def margin_top(self, value):
        self.margin_node.css("margin-top", value)
        self._margin_top = value

    @property
    def border(self):
        return self._border

    @border.setter
    def border(self, value):
        self.break_container.css("border", value)
        self._border = value
```

**form_template.yaml**
```
container:
  type: HtmlTemplate
  properties: {tooltip: '', background: '', foreground: '', border: '', visible: true,
    role: null, html: "<!--\nBased on the snippet at https://anvil.works/forum/t/plots-in-pdf-being-divided-between-two-pages/7774/5\n\
      -->\n<div class=\"break-container\" style=\"overflow: hidden;\">\n  <div style=\"\
      page-break-after:always;\"/>\n  <div class=\"margin-element\" style=\"margin-top:\
      \ 0px;\"/>\n</div>\n\n<style>\n  .break-container {\n    border: 1px solid grey;\n\
      \  }\n  @media print {\n    .break-container {\n      border: none !important;\n\
      \    }\n  }\n</style>"}
components: []
is_package: true
custom_component: true
properties:
- {name: margin_top, type: number, default_value: 0, default_binding_prop: true, description: Use to adjust whitespace at the top of each page in the generated pdf. This is an optional property and defaults to 0. Can be positive or negative. Negative numbers will reduce whitespace in the generated pdf.}
- {name: border, type: string, default_value: 1px solid gray, description: Use to set the style in the form. Useful when the PDF has many grey horizontal lines and the default border style would be confusing. Has no effect on the printed PDF.}
```

s-cork: No need to post the code - the yaml is difficult to read - but that's just because of the way it's formatted.

**To update a PR** - just push another commit to your fork and this PR will update automatically.



Given that when you change the border you probably want to change the border for all PageBreak elements, I think that it might be better to add documentation as to how to change a PageBreak border with an adjustment to `theme.css`. Since this will affect all PageBreak elements both in the design view and the displayed app.


If we change the css in the designer html to be:
```html
<style>
  .break-container {
    border: 1px solid grey;
  }
  @media print {
    .break-container {
      border: none !important;
    }
  }
</style>
```

And then to customize theme.css - you can do something like

```css
  .break-container {
    border: 1px solid red !important;
  }
```

What do you think?

client_code/PageBreak/__init__.py
```diff
@@ -15,8 +15,11 @@
 class PageBreak(PageBreakTemplate):
     def __init__(self, margin_top=0, border="1px solid grey", **properties):
         self.margin_node = _S(anvil.js.get_dom_node(self)).find(".margin-element")
+        self.break_container = _S(anvil.js.get_dom_node(self)).find(".break-container")
```
s-cork, line 18:
```suggestion
        dom_node = _S(anvil.js.get_dom_node(self))
        self.margin_node = dom_node.find(".margin-element")
        self.break_container = dom_node.find(".break-container")
```

docs/guides/components/page_break.rst
```diff
@@ -2,6 +2,21 @@ PageBreak
 =========
 For use in forms which are rendered to PDF to indicate that a page break is required.
 
-The ``margin_top`` property is optional. In the generated pdf you may want more or less white space at the top of the page.
+The optional ``margin_top`` property  allows to change the amount of white space at the top of the page.
```
s-cork, line 5:
```suggestion
The optional ``margin_top`` property  changes the amount of white space at the top of the page.
```

client_code/PageBreak/form_template.yaml
stenci: I agree that most of the times adding that class to theme.css will do the job, even better than using the `border` property, because the change is visible in the IDE too. If I had known about it I wouldn't have bothered adding the property.

I hope I understood your comments: I kept the `border` property, added a the description of both the `border` property and the theme.css customization to the documentation, committed and pushed... and it worked!

docs/guides/components/page_break.rst
```diff
@@ -2,6 +2,21 @@ PageBreak
...
 You can set the ``margin_top`` property to a positive/negative number to adjust the whitespace.
-Most of the time this is unnecessary. This won't have any effect on the designer only the generated pdf.
+Most of the time this is unnecessary. This won't have any effect on the designer, only the generated PDF.
+
+The optional ``border`` property allows to define the style of the component in the IDE.
```
s-cork, line 9: suggestion:

```text
The optional ``border`` property defines the style of the PageBreak.
```

s-cork (changes requested):
lgtm - some code suggestions.

s-cork:
@stenci - are you happy to resolve the suggestions above with another commit?
(I'm also happy to resolve them).

docs/guides/components/page_break.rst
stenci: I think I have addressed all your points.

I also have addressed https://github.com/anvilistas/anvil-extras/issues/142, so I could test it.

s-cork:
Perfect - thanks for fixing #142 - merging

PR #149, opened 2021-09-17
Hide MultiSelect drop menu when component hidden

### LINKED ISSUE #144: Multiselect dropdown remains visible
If it is not specifically dismissed, the multiselect dropdown remains visible after the underlying form is hidden:
MultiSelect Bug

hugetim:
Attempts to resolve #144

### DIFF
--- a/client_code/MultiSelectDropDown/__init__.py
+++ b/client_code/MultiSelectDropDown/__init__.py
@@ -100,8 +100,19 @@ class MultiSelectDropDown(MultiSelectDropDownTemplate):
 
         self._el.selectpicker()
         self._el.on("changed.bs.select", self.change)
-        self._el.on("shown.bs.select", self.shown)
-        self._el.on("hidden.bs.select", self.hidden)
+
+        self._menu_visible = False
+
+        def shown(*e):
+            self._menu_visible = True
+
+        self._el.on("shown.bs.select", shown)
+
+        def hidden(*e):
+            self._menu_visible = False
+
+        self._el.on("hidden.bs.select", hidden)
+
         self._init = True
 
     ##### PROPERTIES #####
@@ -171,13 +182,7 @@ class MultiSelectDropDown(MultiSelectDropDownTemplate):
     def change(self, *e):
         return self.raise_event("change")
 
-    def shown(self, *e):
-        self._menu_shown = True
-
-    def hidden(self, *e):
-        self._menu_shown = False
-
-    def form_hide(self, **event_args):
+    def _form_hide(self, **event_args):
         """This method is called when the HTML panel is removed from the screen"""
         if self._menu_shown:
             self._el.selectpicker("toggle")
--- a/client_code/MultiSelectDropDown/__init__.py
+++ b/client_code/MultiSelectDropDown/__init__.py
@@ -100,6 +100,8 @@ class MultiSelectDropDown(MultiSelectDropDownTemplate):
 
         self._el.selectpicker()
         self._el.on("changed.bs.select", self.change)
+        self._el.on("shown.bs.select", self.shown)
+        self._el.on("hidden.bs.select", self.hidden)
         self._init = True
 
     ##### PROPERTIES #####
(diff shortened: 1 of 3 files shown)


client_code/MultiSelectDropDown/__init__.py
```diff
@@ -169,6 +171,17 @@ def selected(self, values):
...
+        self._menu_shown = True
+
+    def hidden(self, *e):
+        self._menu_shown = False
+
+    def form_hide(self, **event_args):
+        """This method is called when the HTML panel is removed from the screen"""
+        if self._menu_shown:
+            self._el.selectpicker("toggle")
+
```
s-cork, line 184: did you try `self._el.selectpicker('hide')`?
that might mean we don't need to worry about shown and hidden methods/events.

re that bug for `form_hide` - it's been like that for a long time so i'll check with the team and see if that's intentional or not. If it is a bug then when that's fixed this should work.


But if isn't a bug - I'm not sure we need to worry too much about the visible property affecting the menu.
Potentially we could have `hide_menu()` method that can be called. (and a `show_menu` method). 

Or - If we really want the `visible` property to work then we can override the property.

```python
@property
def visible(self):
    return super().visible
    # or _HtmlPanel.visible.__get__(self)

@visible.setter(self, value):
    if not value:
        self._el.selectpicker('hide')
    # we can't user super see:
    # https://stackoverflow.com/questions/13595607/using-super-in-a-propertys-setter-method-when-using-the-property-decorator-r
    _HtmlPanel.visible.__set__(self, value)
```

hugetim:
I did try the "hide" method. Alas, it only hides the base "select"
component, not the drop-down menu.

---
typed with thumbs

client_code/MultiSelectDropDown/__init__.py
```diff
@@ -169,6 +171,17 @@ def selected(self, values):
     def change(self, *e):
         return self.raise_event("change")
 
+    def shown(self, *e):
+        self._menu_shown = True
+
+    def hidden(self, *e):
+        self._menu_shown = False
```
s-cork, line 178: I don't think we want shown and hidden to be public method here.

Consider creating them in the init method
```python
        ...
        self._menu_visible = False

        def shown(*e):
            self._menu_visible = True
        self._el.on("shown.bs.select", shown)

        def hidden(*e):
            self._menu_visible = False
        self._el.on("hidden.bs.select", hidden)

```

client_code/MultiSelectDropDown/__init__.py
hugetim: I did try the "hide" method. Alas, it only hides the base "select" component, not the drop-down menu.

I like the idea of `hide_menu()` and `show_menu()` if the 'hide' event thing is not a bug.

If it is a bug and it gets fixed, I'm a little concerned that the 'hide' event of the HTMLTemplate might be shadowed by the 'hide' event of the custom component so that my current code wouldn't work anyway--though maybe that shadowing is only an issue for properties, not events.

On the `visible` property, I don't understand how your proposed code (or what you have in there for `visible` currently) works. I can't wrap my mind around what it means to be doing these things with `_HtmlPanel` as a class, as opposed to an instance.

I'm going to be offline for the next few days. Don't hesitate to take this over if that's easier. Otherwise, I'll try to finish up late next week.

s-cork: No rush.

---

Yeah it took me ages to understand how functions/methods/classmethods/staticmethods/properties work in Python.
It's a path of becoming really familiar with how Python gets an attribute when you ask for one AND what you're asking for is a descriptor.
A descriptor behaves differently depending on whether the asker is the class or the instance.
(classmethods, functions, staticmethods, properties are all descriptors)
That's a deep rabbit hole to dive into.
This is a nice place to look: https://docs.python.org/3/howto/descriptor.html#pure-python-equivalents

s-cork: Also with the hide event - there's no issue there.
When we added `add_event_handler()` this solved the problem for a custom component using its own event handler and a user defining an event handler for that same event.

Under the hood we call `add_event_handler()` when you define an event handler in the design view.
This means that each event handler gets called in turn - i.e. it won't get overridden by the user.

(On my today is to expose show/hide events for all components now that this isn't an issue)

s-cork:
I think this solves the issue from #144 

`hide_menu()` and `show_menu()` are nice to haves that can be part of a separate feature branch.

The visible property suggestions can also wait since they don't solve the issue of a container changing its visible property.

Merging

PR #153, opened 2021-09-22
Use js.import_from in a few places

s-cork:
A new anvil.js feature which simplifies a couple of javascript library imports for us.
I've identified a few places where we can import modern javascript modules into python using it.

Also removed `from anvil import *` in the slider component which didn't seem to be being used at all 🤷 

(Checked all working on the demo app and ran the modules as main)


meatballs:
Nice! That makes life a little simpler once again.

PR #161, opened 2021-10-02
Some initial guidance for creating designer components

s-cork:
some internal docs both as a reminder for me about what I was trying to do.
And for others brave enough to try.


meatballs:
Nice! Thank you

PR #165, opened 2021-10-03
Add initial pivot table implementation

meatballs:
- [ ] Add working python component
- [ ] Add designer version
- [ ] Write docs


meatballs:
@s-cork Your thoughts on how I've done the initial component would be much appreciated before I attempt to tackle the designer side. (At the moment, I've not bothered with any dynamic properties. That's a conscious decision for now).

client_code/Pivot/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+        "values": "vals",
+        "aggregator": "aggregatorName",
+    }
+
+    def __init__(self, **properties):
+        self.pivot_options = {
+            option: properties[option] for option in self.option_names
+        }
+        spacer = anvil.Spacer()
+        self.node_id = helpers._get_dom_node_id(spacer)
```
s-cork, line 48: These methods won't be (easily) available in the Designer.
consider using a div in your HTML with an identifiable class rather than a spacer.
then use

```
dom_node = get_dom_node(self)
self.pivot_node = dom_node.querySelector(".anvil-extras-pivot")
```

This will make it much easier to do a design version

client_code/Pivot/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+        self.add_component(spacer)
+        self.init_components(**properties)
+
+    def _init_pivot(self):
+        options = {
+            value: self.pivot_options[key] for key, value in self.option_names.items()
+        }
+        _jquery(f"#{self.node_id}").pivotUI(self.items, options)
```
s-cork, line 56: If you go with the `self.pivot_node` - then no need for a query selector just pass `self.pivot_node` to `_jquery()`

s-cork (commented):
Just a couple of comments thinking ahead.

You may also want to use `update` over `setProp` because it looks like you might need to destroy and recreate the pivot table whenever an attribute changes.

the init method will look something like

```typescript
DesignerPivot {
    static init() {
        super.init(".anvil-extras-pivot")
    }
}

File: client_code/Pivot/__init__.py (hunk 1 above)
Comment (meatballs) [reply client_code/Pivot/__init__.py:48]: and how would that cope with more than one instance on a form?

Comment (meatballs) [reply client_code/Pivot/__init__.py:48]: Perhaps use jQuery to add the div with a unique id?

Comment (s-cork) [reply client_code/Pivot/__init__.py:48]: we don't need to worry because when we do `dom_node.querySelector` we're only searching in our child elements.

Comment (meatballs) [reply client_code/Pivot/__init__.py:48]: ahhhh! of course!

File: client_code/Pivot/__init__.py [hunk 3]
```diff
@@ -0,0 +1,59 @@
...
+
+__version__ = "1.6.0"
+
+pivottable_version = "2.23.0"
+jqueryui_version = "1.11.4"
+
+prefix = "https://cdnjs.cloudflare.com/ajax/libs"
+dependencies = [
+    f"{prefix}/pivottable/{pivottable_version}/pivot.min.css",
+    f"{prefix}/jqueryui/{jqueryui_version}/jquery-ui.min.js",
```
s-cork, line 21: I guess it could also be worth having a separate check for a clash with jquery-ui.
I can see someone else adding this in their native libraries for a different component?

s-cork (approved):
lgtm - merge away

PR #214, opened 2021-11-23
add a lazy module loader and load marshmallow using it

### LINKED ISSUE #212: Serialisation module on the free plan
**Describe the bug**
At the moment the development branch breaks free user accounts for server calls
All server modules are loaded on a server call including the serialisation module
On the free plan marshmallow is not available which will throw an error

**To Reproduce**
- depend on the development branch
- switch to python basic 
- call a server function

**Additional context**
We should probably lazy load marshmallow anyway for performance reasons.
(given that the serialisation module will be imported for each server call)

Assuming @meatballs will want to take this.

s-cork:
close #212

### DIFF
--- a/.github/workflows/check_codebase.yml
+++ b/.github/workflows/check_codebase.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6]
+        python-version: [3.7]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


s-cork:
fixes the immediate issue - merging

meatballs:
Nice job!

s-cork:
@meatballs feel free to clean anything that isn't how you'd write it.
I tried to make the pr without moving too many flowers around the garden so to speak

meatballs:
All looks perfectly good to me

PR #244, opened 2021-11-28
If load_from_cache is false

### LINKED ISSUE #243: Possible bug in routing.set_url_hash when load_from_cache=False
I use code like this as part of my login sequence:

```
      anvil.users.login_with_form(allow_cancel=True)
      routing.set_url_hash(self.tag, load_from_cache=False, replace_current_url=True) 
```

The problem comes when the form the user is currently on is the same one referenced by self.tag.  Given that I'm passing load_from_cache=False, I expected the form to be reloaded, with new calls to init and form_show.  

Instead the form simply remained visible, with no calls to any function on the form.

The following code fixed the issue and caused the behavior I expected:

```
      anvil.users.login_with_form(allow_cancel=True)
      routing.remove_from_cache(self.tag)
      routing.set_url_hash(self.tag, load_from_cache=False, replace_current_url=True) 
```

Posting here in case this is a bug in the routing library that needs addressed.

jshaffstall:
remove the page from cache before checking to see if we can get by with doing nothing.   This addresses the issue described in https://github.com/anvilistas/anvil-extras/issues/243

Note that I do not have a testing environment setup to test this! 

Close #243

### DIFF
--- a/client_code/routing/_routing.py
+++ b/client_code/routing/_routing.py
@@ -515,13 +515,13 @@ def set_url_hash(
         url_hash, url_pattern=url_pattern, url_dict=url_dict
     )
 
-    if url_hash == get_url_hash() and url_hash in _cache and _current_form is not None:
-        return  # should not continue if url_hash is identical to the addressbar hash!
-        # but do continue if the url_hash is not in the cache i.e it was manually removed
-
     # remove from cache
     if not load_from_cache:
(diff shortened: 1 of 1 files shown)


s-cork (approved):

s-cork:
@jshaffstall i forgot to request an update to the change log. Would you mind adding a pr for that?

PR #269, opened 2022-02-10
popvers: Support adjusting changing the container and scroll behaviour

### LINKED ISSUE #268: Popover disappearing on page scroll even with auto_dismiss=False
**Describe the bug**

A popover created with auto_dismiss=False disappears when the page scrolls.  This creates a problem when it's the popover itself that is extending beyond the bottom of the page and the user needs to scroll to read it all.

**To Reproduce**

Using this app: https://anvil.works/build#clone:YV6TXJX6VEUJGJD7=LE4R7H7QBXMYW7QHMAJPCZRU

1) Run the app and size the window so that vertical scrollbars are present
2) Click any button to show a popover
3) Scroll the window and the popover disappears

**Expected behavior**

With auto_dismiss=False, I'd expect the popover to remain showing and to scroll with the rest of the page, maintaining its relative position to the other components.  This would allow the user to scroll to read the popover content if needed.

No.

s-cork:
close #268

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# Unreleased
+
+## New Features
+* Popovers - supports changing the default container to something other than `"body"`
+  add ``dismiss_on_scroll()`` and ``set_default_container()`` methods
+  https://github.com/anvilistas/anvil-extras/pull/268
+
+
 # v1.9.0 27-Jan-2022
 
 ## New Features
--- a/client_code/popover.py
+++ b/client_code/popover.py
@@ -23,7 +23,14 @@ from anvil.js.window import jQuery as _S
 
 __version__ = "1.9.0"
 
-__all__ = ["popover", "pop", "dismiss_on_outside_click", "set_default_max_width"]
+__all__ = [
+    "popover",
+    "pop",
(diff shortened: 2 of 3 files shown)


jshaffstall:
What about a helper to automatically add the popover container div to a given spot, so that the page template doesn't need manually modified?  Something like `add_popover_container('.content', 'popover-container')`  I always worry about upgrading manual edits to the built-in HTML and CSS files in the event that new versions are released.

s-cork:
Yeah. My instinct says something like that might be better in the docs.

Have you tried it with this update to see if it works? (Just set the anvil_extras to the development version)

If your code snippet didn't work - which might happen if the popovers try to add themselves before you've appended the container to h the document - then this variation might help. 

```python
popover_container = document.createElement("div")
popover_container.style...

popover.set_default_container(popover_container)
# should work 
```

And then append the popover to the content in the form show.

jshaffstall:
Yeah, that variation worked well.  That way the popovers can be created anywhere, and appending the container can still happen in the form_show.

It also works great when that setup is done in the main routing form, and then page forms can use popovers without change.

s-cork:
@jshaffstall  I've updated the documentation in this PR #270

PR #274, opened 2022-02-24
Quill

### LINKED ISSUE #273: Quill set_html_dangerously(html)
The Quill editor allows you to include images which are returned in the html string as . 

When you set such an html string to a quill, the image is dropped because the html has been sanitized into safe html.

**Describe the solution you'd like**

Could we include a method that does not sanitize the html, and thus keeps the images?

```
def set_html_dangerously(self, html):
        """set the content to html, with no conversion to safe html """
        self._quill.root.innerHTML = html
```

If you agree to the change, I was hoping you could just paste the code as-is or edited, and we could avoid the PR, if that's ok.

s-cork:
close #273


meatballs:
This all looks good to me. It just needs the merge conflicts resolving.

PR #281, opened 2022-03-11
routing: complete refactor and implement templates

s-cork:
Issues:
- the code was a bit of a mess - I wrote it a while ago while I was early in my jedi training
- a common issue is top level forms - routing only supports one

Refactor:
- no longer does @route and @main_router return subclasses of the cls being decorated
- moved some things around e.g. some public api stuff now just lives in `__init__`
- I think it might actually be somewhat readable now - the `routing._router.navigation()` method is the core of i

Templates:
- `@main_router` is equivaluent to `@template(path="", priority=0, condition=None)`
i.e. the main_router is just a default template so a fall back.
- whenever routing navigation is triggered we check the template is valid
  - the current `url_hash` must start with the template path
  - the condition must return `True` (this should be a cheap check)
- templates are checked in priority order
- templates are not cached so will load a fresh each time
- if navigating requires a template change we:
  - halt the navigation
  - wait for the template to load
  - re-navigate

NavigationExit:
- if thrown inside the template's `on_navigation` callback, the current navigation will be halted
- This is useful internally but also if you want to have a login form that browser back and forward does nothing
- and you don't want routing to change your content panel

```python
@routing.template("", priority=2, condition=lambda: Globals.user is None)
class LoginForm(LoginFormTemplate):
    def on_navigation(self, **url_args):
        raise routing.NavigationExit # don't update my content panel
    def login_btn_click(self, **event_args):
        user = anvil.users.login_with_form()
        if user is not None:
            Globals.users = user
            routing.set_url_hash("", replace_current_url=True)
```

**Breaking Change**
- Deprecate and throw a runtime error for `routing.load_form()`
  - it was a bit of a beast to maintain and the same functionality can be achieved with `routing.set_url_hash()`

**Tested**
- I tested this branch on the hash routing example and it works as before

TODO:
- [x] update changelog
- [x] document these changes

_(but don't let that stop you from reviewing the code)_


meatballs:
This all looks much cleaner - nice job!

s-cork:
I'm declaring this ready for review

client_code/routing/__init__.py
```diff
@@ -7,25 +7,194 @@
...
-    clear_cache,
-    error_form,
-    get_cache,
+from anvil.js import window as _w
+
+from . import _navigation
+from . import _router as _r
+from ._decorators import error_form, route, template
+from ._logging import log as _log
+from ._logging import logger
```
meatballs, line 16: I think the _logging module needs moving into utils (and incorporating into the _timed module too)

s-cork: good idea

s-cork: I'll make a separate PR for that

client_code/routing/__init__.py
```diff
@@ -7,25 +7,194 @@
...
-    add_to_cache,
-    clear_cache,
-    error_form,
-    get_cache,
+from anvil.js import window as _w
+
+from . import _navigation
+from . import _router as _r
+from ._decorators import error_form, route, template
+from ._logging import log as _log
```
meatballs, line 15: I can't see a `log` attribute in the `._logging` module. What's this importing?

s-cork: It should be in the diff at the end of the _logging module

It's just a function to send a callable to logger.print, which just means we don't need to realise the f-string if we're not in debug mode. It's like a tiny performance win.

meatballs: ah. stupid of me not to check if it's something new in here!

meatballs (approved):

PR #282, opened 2022-03-12
Logging

s-cork:
This is an idea for a client side logging module based on the one that was in routing

Inspiration from python logging and also from https://github.com/mtayler/simplelogger/blob/master/simplelogger/simplelogger.py
_(I don't know if it's close enough to need to mention the license)_

I tried to make it small - so didn't bother with properties or methods like `set_level()`.
It doesn't necessarily need all the logging levels

Questions:
- ~is the `__setattr__` api too cute?~
  ~`logger.debug = True`~
  ~I put this in because I have it documented in the routing module as the way to update logging~
  ~plus it's very convenient~
  But easy to remove and make the routing logger still backwards compatible 
- ~I added the `logger.print()` api because it's convenient to just go through all print statements and prepend them with `logger.`~
- Where should it live? utils or its own top level module?
 

TODO:
- [x] document
- [x] change log

Will document after we finalise the api


s-cork:
I decided that the api
```python
logger.debug = True
```

passed the cute test and so had to go. I've included it for the routing module for backwards compatibility.

I've added a `TimerLogger` which is an API i'd like to have
The API might be trying to do too much. I often want to time bits of a function and end up writing too much code

I'm also not convinced that the `@timed` decorator ever supported a custom logger with a custom log level
Have adjusted accordingly.

I think that `logging` might warrant being a top level module rather than existing in `utils`.

meatballs:
I like the look of this. Some nice touches and I agree with your choices.

I also reckon it should be a top level module.

Issue #277, opened 2022-02-27
Problem with published branch


s-cork:
**Describe the bug**
Published branch does not point to the 1.9.0 release

**To Reproduce**
```python
from anvil_extras import serialisation
# ImportError
```

**Expected behaviour**
serialisation was part of the latest release so something is up

**Additional context**
flagged in #275 

**I am considering making a PR for this issue**
No

hugetim:
I can confirm this is still an issue. I was still seeing [this bug](https://github.com/anvilistas/anvil-extras/issues/271) when set to the Published branch, but it goes away when I explicitly set the version to 1.9.0 or 2.0.0.

Actually, it may be best to leave the Published branch pointing to 1.9.0 to avoid breaking changes for those with the dependency set to "Published." But then we should just change the [docs](https://github.com/anvilistas/anvil-extras/blob/main/docs/guides/installation.rst#install-as-a-third-party-dependency) to note that.

meatballs:
I'll fix the published branch manually and then add something to the github action to do so from then on.

hugetim:
Last thought: if you're at all considering leaving 'Published' at v1.9.0, you could create a new Published2 branch that updates automatically with new v2.0.x releases.

PR #286, opened 2022-03-15
Update installation.rst: changes to versioning

hugetim:
Addresses #277 (assuming that the 'Published' branch will be set to v.1.9.0 permanently)


s-cork:
Yeap I agree. If people are using anvil extras and are concerned about changes we make they really ought to pin to a version and then, follow our releases and when we make a release, test their app with the bumped version before bumping it themselves. 

A separate PR fixing spelling and wording would be great. 

Something mentioning the preference for pinning to a version rather than using the Production branch. Also mentioning that the versioned branches will never change, whereas the production branch is subject to change with each release and may occasionally include breaking changes. (If we bump a major release it likely includes a breaking change). Any breaking changes will be included in the changelog and posted in the release notes in anvil forum. 

Thanks @hugetim

PR #294, opened 2022-03-22
Routing

### LINKED ISSUE #293: Possible issue with routing templates
**Describe the bug**

This may just be my misunderstanding how things should work.  With the new routing template system, the templates are containers for anything matching the template's URL prefix.  

I have two templates:

```
@routing.default_template
class RouterPublic(RouterPublicTemplate):
```

and

```
@routing.template(path="user", priority=1, condition=lambda: Globals.user is not None)
class RouterUser(RouterUserTemplate):
```

The public router includes a form that matches the `''` url pattern, and the user router includes a form that matches the `'user'` pattern.  In my mind those are the home forms for both templates.  

What I'd expected was that if the user not not logged in, the public router would be the template used, and only forms imported in the public router would be able to be shown.  

What I'm seeing is that the public router is shown, but if I modify the URL to go to #user the user home form is shown in the public template.  

This makes sense from a technical perspective, since the routing library has all the forms registered.  But I'd expected the forms to inherit the condition of the enclosing template.  I don't know if that should happen, or is a feature request, or will never happen.

s-cork:
close #293 

@jshaffstall - wanna have a look here?

It implements almost the API we talked about.

The only difference is that a cached form is bound to both the url hash and the name of the current template
Which means that if route form can be used by multiple templates it will be a new route form per template
Implementing that was too much to think about.

It adds the `template` as a possible argument to the `route` decorator
`template` can be a string or an iterable of strings (defaults to None, meaning I can be used by any template).

When we check the cache we check the cache against the url_hash and the current template's name.
When loading a route form if the route url_pattern does not start with the template's path argument, we prepend the path argument to the route form's url_pattern.

```python
@template("accounts", ...)
class AccountsRouter:
    ...

@tempate("admin", ...)
class AdminRouter:
    ...

@route("/forms", template=["AccountsRouter", "AdminRouter"], ...)
class FormsRoute:
    ...
```

---

**Changing _router.launch() in the template init**

I wasn't totally happy with using a `setTimeout` in the template init method.
This makes things asynchronous.

Instead I move the `_router.launch()` to the first show event.
I think is better.
It means that things run synchronously while maintaining the ability for the template to change the url_hash within its init method.

---

**Mitigating against breaking the shape of the underlying `_cache`**

The cache used to look like:
```python
>>> _cache
{"": <HomeForm object>, ...}
```

But now looks like:
```python
>>> _cache
{("", "MainForm"): <HomeForm object>, ...}
```

I've made this as not breaking as possible by implementing the `_cache` object as a subclass of dict
```python
>>> type(_cache)
_Cache
>>> remove_from_cache("")
>>> _cache
{}
# if removing a string the cache will combine the string with the name of the open_form
# equivalent to _cache.pop(("", type(get_open_form()).__name__), None)
```
similarly adding a url_hash to the cache will combine the string with the name of the open form
```python
>>> add_to_cache("", <NotHomeForm object>)
>>> _cache
{("", "MainForm"): <NotHomeForm object>, ...}
```

Technically if someone is really doing some funny stuff with the `_cache` object by doing
```python
>>> cache = routing.get_cache()
# now directly manipulate the cache
```
It could break for them - but they'd have to really try to manipulate the cache object.
And the docs clearly state
`"Adjusting the cache directly may have side effects."`

---

**TODO:**
- [x] document these changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# Unreleased
+
+## Features
+* routing -  a template argument was added to the `@routing.route` decorator.
+  This argument determines which templates a route can be added to.
+  https://github.com/anvilistas/anvil-extras/issues/293
+
+
 # v2.0.1 16-Mar-2022
 
 ## BugFixes:
--- a/client_code/routing/_decorators.py
+++ b/client_code/routing/_decorators.py
@@ -29,6 +29,7 @@ def template(path="", priority=0, condition=None):
 
         def on_show(sender, **e):
             sender.remove_event_handler("show", on_show)
+            # wait till the show event so that this template is the open_form before re-navigating
             _router.launch()
 
         @wraps(cls_init)
--- a/client_code/routing/_decorators.py
+++ b/client_code/routing/_decorators.py
@@ -8,14 +8,13 @@
 from collections import namedtuple
 from functools import wraps
 
-from anvil.js.window import setTimeout
-
 from . import _router
 
 __version__ = "2.0.1"
 
 route_info = namedtuple(
-    "route_info", ["form", "url_pattern", "url_keys", "title", "fwr", "url_parts"]
+    "route_info",
+    ["form", "url_pattern", "url_keys", "title", "fwr", "url_parts", "templates"],
 )
 template_info = namedtuple("template_info", ["form", "path", "condition"])
 
@@ -34,15 +33,20 @@ def template(path="", priority=0, condition=None):
 
         cls_init = cls.__init__
 
+        def on_show(sender, **e):
+            sender.remove_event_handler("show", on_show)
+            _router.launch()
+
         @wraps(cls_init)
         def init_and_route(self, *args, **kws):
+            _router._ready = False
             try:
-                _router._ready = False
                 cls_init(self, *args, **kws)
-                setTimeout(_router.launch)
-                # use set timeout so that if called with open_form
-                # then the main_router is the open form before we try to navigate
-
+                handlers = self.get_event_handlers("show")
+                self.set_event_handler("show", on_show)
+                # make us the first show event handler and re-add existing
+                for handler in handlers:
+                    self.add_event_handler("show", handler)
             finally:
                 _router.ready = True
 
@@ -59,12 +63,20 @@ class route:
     @routing.route(url_pattern=str,url_keys=List[str], title=str)
     """
 
-    def __init__(self, url_pattern="", url_keys=[], title=None, full_width_row=False):
+    def __init__(
+        self,
+        url_pattern="",
+        url_keys=[],
+        title=None,
+        full_width_row=False,
+        template=None,
+    ):
         self.url_pattern = url_pattern
         self.url_keys = url_keys
         self.title = title
         self.fwr = full_width_row
         self.url_parts = []
+        self.templates = template
 
     def as_dynamic_var(self, part):
         if len(part) > 1 and part[0] == "{" and part[-1] == "}":
@@ -90,6 +102,10 @@ class route:
         self.url_parts = [
             self.as_dynamic_var(part) for part in self.url_pattern.split("/")
         ]
+        if self.templates is None or isinstance(self.templates, str):
+            self.templates = (self.templates,)
+        else:
+            self.templates = tuple(self.templates)
 
     def __call__(self, cls):
         self.validate_args(cls)
--- a/client_code/routing/_decorators.py
+++ b/client_code/routing/_decorators.py
@@ -5,18 +5,12 @@
 #
 # This software is published at https://github.com/anvilistas/anvil-extras
 
-from collections import namedtuple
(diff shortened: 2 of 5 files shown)


jshaffstall:
>Which means that if route form can be used by multiple templates it will be a new route form per template

I think that's fine.  People can work with a single template modifying its sidebar links manually if they want those to change and keep the same instance of the form.

The api looks good, and should solve the original issue I was seeing.  Once it hits the development version of extras I'll play with it.

s-cork:
note - I think we should leave this on the development branch for a little bit - let @jshaffstall  play with it before we commit to a `v2.1.0`.

jshaffstall:
>Note that the cached version of the Foo form will be added to either templates.
>If you don't want to use a cached version for different templates, you should use multiple decorators

That sounds different from what you'd said earlier:

>Which means that if route form can be used by multiple templates it will be a new route form per template

Is that because you worked out how to use the same cached form in multiple templates?  Or is the documentation off?

s-cork:
@jshaffstall  see https://github.com/anvilistas/anvil-extras/pull/294#issuecomment-1075208866

jshaffstall:
Ah, missed that, that's excellent!

s-cork:
@jshaffstall - now that's merged let us know how you get on with the development branch

jshaffstall:
I have what I think is a simple enough app using templates and routes that belong to templates, and am getting an error.  

These are the templates:

```
@routing.default_template
class RouterPublic(RouterPublicTemplate):

@routing.template(path="user", priority=1, condition=lambda: Globals.user is not None)
class RouterUser(RouterUserTemplate):
```

And these are the routes:

```
@routing.route('', template='RouterPublic')
class Home(HomeTemplate):

@routing.route('', template='RouterUser')
class Home(HomeTemplate):
```

When I start the app and it shows the public home page, if I edit the url to add '#user' it gives me an error:

```
LookupError: 'user' does not exist
at app/anvil_extras/routing/_router.py:237
called from app/anvil_extras/routing/_router.py:270
called from app/anvil_extras/routing/_router.py:196
called from app/anvil_extras/routing/_router.py:120
called from app/anvil_extras/routing/_router.py:120
called from app/anvil_extras/routing/_navigation.py:54
```

s-cork:
did you import it? 
clone link?

s-cork:
@jshaffstall I tried to reproduce but it works for me
https://anvil.works/build#clone:NIUIZTVBDB6SUGKP=QA3PYOD5MOTX5ZLSLB7UTQUM

If you turn on the `routing.logger.debug = True` you should see a helpful message about a potential missing import

jshaffstall:
Sorry, meant to include a clone link, but got called away: https://anvil.works/build#clone:KFF6KEV4NDMJJK5U=JI5EORBERTEDFD7AG4O5WTGW

I do have the user form imported in RouterUser.  The issue is that routing is looking in RouterPublic:

```
#routing: no route form with:
	url_pattern='user'
	url_keys=[]
	template='RouterPublic'
```

RouterPublic is the currently showing template, but when I edit the URL to be `#user` that should trigger a load of the RouterUser router, right?

s-cork:
I had a look @jshaffstall, when I login successfully I get navigated correctly. 

When I fail to login I get the error. 

Which makes sense because the template RouterUser doesn't take over if the login attempt fails because of its condition argument.

jshaffstall:
Yeah, I get it now.  The condition puts RouterPublic as the template because of not being logged in, then tries to open `#user` in that template, which doesn't exist.  

The behavior I was trying to get would be the same as from before templates, when I'd put something like this into the `#user` page:

```
user = anvil.users.get_user()

if not user:
  routing.set_url_hash('')
```

But now the `if not user` check is taken care of by the template, and the `#user` page can't say where to redirect when the condition fails since it's never invoked.

This seems like it'll be a common occurrence with templates whose conditions fail.  The URL structure under that template will attempt to apply itself to the next template whose condition passes, and won't be found.

Would it be possible to specify a fallback URL for those templates that have conditions, for use when the condition fails?

s-cork:
Since the condition is a contract about when the template is invoked, I'm not sure what a practical API is for what you're wanting to happen?

I can't see an API that doesn't involve moving the conditional logic out of the condition attribute and adding it somewhere else. 

I think you want a redirect to happen but to be taken care of by the template decorator?


Like
`@template("user", priority=1, condition=..., redirect="")`

Which might be saying, ok if my path is valid and the condition is false invoke the redirect. 

Thoughts?

jshaffstall:
That's what I was thinking in terms of, yes.  

Given the current functionality, the only place these sorts of redirects could happen is in the error form, but that's an odd place to do redirects, and it centralizes information about which templates redirect where.  It'd be hard to distinguish between a failed template condition and a legitimate 404 situation.

I think having the information about the fallback redirect on the template decorator itself makes more sense.   Or for there to be some other way of doing a redirect from a page that never gets loaded.

PR #297, opened 2022-03-26
Routing - add a redirect argument to templates

s-cork:
@jshaffstall - take a look - what do you think?

There's a choice about how the redirect should work
I went with 

```python
            set_url_hash(
                redirect, replace_current_url=True, set_in_history=False, redirect=True
            )
```
But I could see an argument for `set_in_history=True`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 * routing -  a template argument was added to the `@routing.route` decorator.
   This argument determines which templates a route can be added to.
   https://github.com/anvilistas/anvil-extras/issues/293
+* routing - templates can take a redirect url_hash argument,
+  if the condition() fails on a template routing will redirect to this url_hash if present
+  https://github.com/anvilistas/anvil-extras/issues/295
 
 
 # v2.0.1 16-Mar-2022
--- a/client_code/routing/_decorators.py
+++ b/client_code/routing/_decorators.py
@@ -13,16 +13,18 @@ from ._utils import RouteInfo, TemplateInfo
 __version__ = "2.0.1"
 
 
-def template(path="", priority=0, condition=None):
+def template(path="", priority=0, condition=None, redirect=None):
     if not isinstance(path, str):
         raise TypeError("the first argument to template must be a str")
     if not isinstance(priority, int):
         raise TypeError("the template priority must be an int")
     if condition is not None and not callable(condition):
         raise TypeError("the condition must be None or a callable")
+    if redirect is not None and not isinstance(redirect, str):
+        raise TypeError("redirect must be set to a str")
 
     def template_wrapper(cls):
-        info = TemplateInfo(cls, path, condition)
+        info = TemplateInfo(cls, path, condition, redirect)
         _router.add_template_info(cls, priority, info)
 
         cls_init = cls.__init__
--- a/client_code/routing/_router.py
+++ b/client_code/routing/_router.py
@@ -156,13 +156,23 @@ def load_template(url_hash):
 
     logger.debug("Checking routing templates")
     for template_info in chain.from_iterable(_ordered_templates.values()):
-        cls, path, condition = template_info
+        cls, path, condition, redirect = template_info
         if not url_hash.startswith(path):
             continue
         if condition is None:
             break
         elif condition():
             break
+        elif redirect is not None:
+            from . import set_url_hash
+
+            logger.debug(
+                f"{cls.__name__!r} routing template url valid, condition() failed, invoking redirect={redirect!r}"
+            )
+            set_url_hash(
+                redirect, replace_current_url=True, set_in_history=False, redirect=True
+            )
+            raise NavigationExit
     else:
         load_error_or_raise(f"No template for {url_hash!r}")
     if current_cls is cls:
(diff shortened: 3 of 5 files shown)


jshaffstall:
I'm concerned about the template redirect hiding valid 404 results.  

Before the redirect url was available valid URLs would get marked as not being valid (false positives).  The redirect solves that.

With the redirect invalid URLs will be redirected instead of being marked as not valid (false negatives).  This seems like a problem.  If `#user/account` is a valid page on my site, but they don't have permissions to see it, the redirect should happen.  But if they go to `#user/foo`, which is not a page on my site, I'd expect the error form to trigger.  

Detecting that, though, would require going through every page registered to the template in the `elif redirect is not None:` block and redirecting only if the URL matched a page.  

Unless I'm overthinking this?

jshaffstall:
I guess the behavior is different depending on the web app.  Blackboard, a course management system, does the opposite,  you get a login page for valid protected URLs, and a 404 page for invalid protected URLs.  The more I think about it, I guess it doesn't really matter, so the current behavior works fine.   As long as we avoid marking valid URLs as invalid, which the redirect does. 

Another thing to think about.  Some web apps, when you try to go to a protected URL, will redirect to the login page.  Then when you login they take you to the original protected URL you were trying to go to.  Before templates with conditions that would be managed in forms.   

With conditional templates, is there a way to get to the original URL hash before the redirect?

jshaffstall:
It strikes me that if there were a way to have wildcard url patterns, everything we're trying to do with redirects could be handled with a high priority login template that used wildcard url patterns to match all protected areas with a condition of the user not being logged in, as long as the login template could get to the original URL to redirect to after login.  Other sorts of protected areas could have a similar gatekeeping template.

Not trying to advocate for one or the other approach; just trying to workout if redirects are the easiest way to solve the issue.

s-cork:
I've thought about wild card patterns - then stopped short when trying to implement them.

Do you have a pattern in mind about how you'd like to use it?

There are a couple of ways i'd currently implement a catch all for a login template that could redirect to the original url after login.

jshaffstall:
For what I was suggesting, just * would be enough, as long as multiple template routes could be specified

```
@routing.template(path=["admin/*", "account/*"], priority=20, condition=lambda: Globals.user is None)
class RouterLogin(RouterLoginTemplate):
```

That template could then redirect to the original url after login, or to the home page if the login was cancelled.

s-cork:
That wouldn't need the * no?
Just a way to define multiple paths?

Once that existed how would you envisage the rest of the code. 

Currently I'd do something like. 

```python
@template(path=[...], ...)
class LoginRouter(...):
    def on_navigation(self, url_hash,...):
    self.current = url_hash
        routing.set_url_hash("login", replace_current=True, set_in_history=False, redirect=False)
    raise routing.NavigationExit
    # don't change my content panel 

def login_btn_click(self, **event_args):
    # login user
    routing.set_url_hash(self.current)
```python

**Comment by jshaffstall** [comment]
Yeah, that was a brain dead moment.  Template urls are already an implicit * match at the end.

Given a way to match multiple urls for a template, I think we're good.  That should remove the need for the redirects.  

The same technique should be useful for pages that require higher levels of privilege than just being logged in.  This template would be processed at a lower priority than the login one and can redirect to a suitable page (assuming that setting the url hash inside on_navigation transfers to a new template)  or display an error message:

```
@template(path=[...], condition=lambda: not Globals.user['admin'])
class NotAdminRouter(...):
    def on_navigation(self, url_hash,...):
        routing.set_url_hash("", replace_current=True)
```

**Comment by s-cork** [comment]
Yeap that could work. I'd need to think about redirecting the url_hash in that situation. 

And might be better just to do that type of immediate redirect in the `__init__` method. 

I'd need to make sure it doesn't end up doing the equivalent of `open_form()` within the `__init__` of the current open form. 

i.e how to raise a NavigationExit implicitly when the current navigation get superseded by a new call to set_url_hash. But that shouldn't be too difficult.

**Comment by jshaffstall** [comment]
Yeah, ideally the redirect would happen before the current form shows.  I hadn't thought `__init__` would work because of the limitation on opening another form in there.

It feels odd defining a template form whose purpose is to never get shown.

**Comment by s-cork** [comment]
Yeah it does feel odd. What's the point in making it a form?

That example should work but - it's possible a new api might be better suited here.

You could register a redirect class method.

```python
@redirect(path=[], ...)
def no_admin_redirect():
   routing.set_url_hash(..., )

# or as a class
@redirect(path=[], priority=20...)
class NoAdminRedirect:
    def __init__(self):
        routing.set_url_hash(..., )
```

and a redirect I guess could do nothing? And that would be ok?
Or if a redirect is invoked as part of navigation then we expect the redirect to actually redirect. 

Redirects would just be called along with template depending their priority. 
 
It would essentially be like a template but it would never become the open_form since it's not a form.

jshaffstall:
I like the redirect function/class idea.  That'd work great.

If a redirect function doesn't actually redirect, would routing continue with the next lower priority template?  If so, I think it's fine to not require the redirect to happen.  I'm coming up short trying to work out when that would be useful (everything I think of can just be done as part of the redirect condition), but it wouldn't hurt to allow it.  Maybe someone will come up with a clever use for it. 

Returning the url hash/None/False makes sense, since it's a function designed to redirect, and that saves some typing.

s-cork:
Superseded by #298

PR #298, opened 2022-03-27
Template paths and redirects

s-cork:
@jshaffstall 

attempt number 2.

Maybe you could have a play with it by cloning this version of anvil-extras:
https://anvil.works/build#clone:6ATPRINDTZUXBKSQ=R3XKGGOVQZK6MZVHG36HCTLU

Then changing the dependency to the local copy while we still work out the kinks.

I changed a bunch of logging statements as I was working on it to make it easier to see what was going on.

The api is the one we came up with on the other branch:

```python
@routing.redirect(...) # same args as template
def redirect_no_admin():
    # routing.set_url_hash("login", ...)
    # or return a url_hash
    return "login"

```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,7 @@ repos:
           # these are errors that will be ignored by flake8
           # check out their meaning here
           # https://flake8.pycqa.org/en/latest/user/error-codes.html
-          - "--ignore=E203,E266,E501,W503,F403,F401,E402"
+          - "--ignore=E203,E266,E501,W503,F403,F401,E402,F811"
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.1.13
     hooks:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,14 @@
 # Unreleased
 
 ## Features
-* routing -  a template argument was added to the `@routing.route` decorator.
+* routing - a template argument was added to the `@routing.route` decorator.
   This argument determines which templates a route can be added to.
   https://github.com/anvilistas/anvil-extras/issues/293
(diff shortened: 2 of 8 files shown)


jshaffstall:
Not sure what is wrong with this: https://anvil.works/build#clone:KFF6KEV4NDMJJK5U=JI5EORBERTEDFD7AG4O5WTGW

When I run it under the regular development version of anvil_extras, it shows the public home page when run.  When run under the clone version it does not show the home page.  I've commented out the redirect and login template for now.

s-cork:
oh sugar - that should be fixed now - trying to prevent an infinite loop caused the first loop to not execute!

Fixed - 

if you're git savy you can clone the anvil-extras-dev locally
clone this repo locally
and do some git incantations rather than recloning

jshaffstall:
Uncommented out the redirect and login router.

The login router intercepts correctly when trying to go to any of its routes.  

The first run I got an error about self.current not existing in login_click, but after that one run it worked fine.  Not sure what that was about.

The redirect decorator works great, non-admins are getting redirected to `#user`.

As far as I can tell from the setup I have, it all seems to be working fine.  This technique reduces the number of template conditions needed by quite a bit.  I only have conditions on my login router and my no admin redirect.  By the time it gets to the user or admin routers, I already know the user is logged in and is an admin if they're trying to go to the admin pages.

s-cork:
So we're calling this branch a win.

I'll give it another once over later today
add some docs
and it'll be ready for review

side note - i see you're using/experimenting with Tabulator.
Definitely move to version v2.1.0 if you can - it's recently gone through a rewrite.

jshaffstall:
Yeah, I wrote another sample app that included some pages and sidebar links, and everything worked as expected.  Definitely a win.

s-cork:
@jshaffstall this is now on the development branch
keep us posted on anything you notice in your experiments!

PR #301, opened 2022-03-29
Add client side hashlib implementation

meatballs:
- Add hashlib module
- Add version number to hashlib

### DIFF
--- a/client_code/hashlib.py
+++ b/client_code/hashlib.py
@@ -19,11 +19,11 @@ def digest(algorithm, data):
     ----------
     algorithm : str
         The algorithm to use.
-    data : object
+    data : str or bytes
         The data to digest.
     """
     if not isinstance(data, (bytes, str)):
-        data = str(data).encode("utf-8")
+        raise TypeError("data must be a string or bytes object")
 
     if isinstance(data, str):
         data = data.encode("utf-8")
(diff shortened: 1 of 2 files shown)


client_code/hashlib.py
```diff
@@ -0,0 +1,39 @@
...
+    algorithm : str
+        The algorithm to use.
+    data : object
+        The data to digest.
+    """
+    if not isinstance(data, (bytes, str)):
+        data = str(data).encode("utf-8")
+
+    if isinstance(data, str):
+        data = data.encode("utf-8")
```
s-cork, line 29: why do we allow anything other than a str or bytes here?
just curious - seems odd to support a mutable data structure?

docs/guides/modules/hashlib.rst
```diff
@@ -0,0 +1,21 @@
...
+=======
+A client module that provides several hashing algorithms.
+
+Usage
+-----
+The module provides the functions ``sha1``, ``sha256``, ``sha384`` and ``sha512``. Each
+can be called by passing the data to be hashed and will return a hex string.
+
+The data can be any object. If it is anything other than a string or bytes instance, it
+must have ``__str__`` method available.
```
s-cork, line 11: when you say it must have a `__str__` - all python objects inherit a `__str__` so not sure what the sentence is trying to add?
might be better to say it will be converted to a `str` object before being hashed.

But i'd still consider a `TypeError` for anything other than `str` or `bytes`.

s-cork (approved):
Just a question mark about the data attribute, otherwise looks good and really clean!

meatballs:
I have tested this to ensure it gives the same hash values as the standard library implementation. However, its quite tricky to add those tests properly to the test suite.

PR #302, opened 2022-04-01
implement utils.import_lib and use this for our own lazy utils loader

s-cork:
I also reorganised the `__dir__` to be alphabetical

### DIFF
--- a/client_code/utils/__init__.py
+++ b/client_code/utils/__init__.py
@@ -5,6 +5,7 @@
 #
 # This software is published at https://github.com/anvilistas/anvil-extras
 
+import sys
 from functools import lru_cache
 
 __version__ = "2.0.1"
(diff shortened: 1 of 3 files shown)


client_code/utils/__init__.py
```diff
@@ -13,36 +14,60 @@
...
+        if not package:
+            msg = (
+                "the 'package' argument is required to perform a relative "
+                "import for {!r}"
+            )
+            raise TypeError(msg.format(name))
+        for character in name:
+            if character != ".":
+                break
+            level += 1
```
meatballs, line 43: could this just be:
``` python
level = name.count(".")
```

s-cork: It's just the first two dots in  `"..foo.bar"`

This part is identical to the implementation for `importlib.import_module`

meatballs: Gotcha. I misread it!

PR #311, opened 2022-04-16
Auto Refreshing Proxy

### LINKED ISSUE #250: utils.auto_refreshing - not calling refreshing_data_bindings event when self.item is None during init_components()
As per #249 the form's `refreshing_data_bindings` event is not being called if `self.item` is `None`

Solution has been created and will be put into a PR soon.

s-cork:
Just an idea which I quite like
it's potentially breaking - though I think we can get away with a minor release.

Instead of subclassing from dict - we proxy the original item object.
Changes to the proxied item will call auto_refreshing and also update the original item object.

This means that we can do `auto_refreshing` on any object.
We expect a dictionary or anvil table row, but most sensible substitutes for an `item` would work just fine.

Sending the item to the server will just send the original item.
_(I changed to `lru_cache` because `cache` is not available on the server for when the class is portable)_

referencing #89 
closes #250 

@hugetim, @rhurlbatt - you might be interested in this one since we've talked about it before

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Unreleased
 
+## Notable Change
+* `@auto_refreshing` - the original item will now be proxied rather than copied.
+  Changes to the proxied item will make changes to the original item.
+  https://github.com/anvilistas/anvil-extras/pull/311
+
 ## New Features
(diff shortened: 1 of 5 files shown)


meatballs (approved):
Nice!

Issue #307, opened 2022-04-15
Multiselect does not refresh data bindings on change


erikcw:
**Describe the bug**
When a Multiselect has `selected` bound to a value in `self.item['...']`, it doesn't run `self.refresh_data_bindings()` on change.  Doing so in a change event leads to some sort of infinite loop and crashes the app/browser.

**To Reproduce**
Steps to reproduce the behavior (including a clone link when applicable):

Here is an example to clone: https://anvil.works/build#clone:V7LJDPUIVR4GLXCS=X37GLK3PCQX35FC22C7XWP6Z

**Expected behavior**
Like other Anvil form elements, changes should refresh data bindings.

**I am considering making a PR for this issue**
No

erikcw:
@s-cork This PR is a significant improvement.  However, 2 issues still seem to remain.

1)  It looks like `self.refresh_data_bindings()` still isn't being run `onChange` or `onClose` by default like other anvil components that are using data bindings typically do.  Adding it in a change method works, but is a deviation from Anvil norms.
2)  More importantly, if `selected` is bound to `self.item[...]` in the IDE, and the item has a value on init (`self.item['selected'] = ['a']` ), the state of the multiselect remains empty/null.  I've updated my example above to reflect this.

s-cork:
Do you have an example of a component where you don't need to call `refresh_data_bindings()`?

I think that might be a mistaken expectation. Try the same thing with a TextBox and a Label and you'll still need to call `refresh_data_bindings()` in the TextBox change event. 

That's what `anvil_extras.utils` auto_refreshing is designed to solve. 

Thanks for the update clone. I'll take a look and the second point.

erikcw:
Actually, I think you are right about my mistaken expectation.  I just added the Native Anvil DropDown, and it doesn't `refresh_data_bindings()` automatically.

Let me know your thoughts on the 2nd point.

rhurlbatt:
I suspect there is an edge case issue here with this code in the Dev Branch. Previous code that was working now is not. It seems that the MultiSelect is no longer firing its change event when that event was set in code. I had this to set the change event in the form show which worked before the weekend.

`self.supplier_msdd.set_event_handler('change', self.supplier_msdd_change)`

Now the change event no longer fires. I have rolled back my version to v2.01 and all seems to be working again.

I am not meant to be in the office today so won't get to have a look at this until tomorrow. If needed I can try build a minimal example then.

s-cork:
[...]

PR #313, opened 2022-04-20
support selectAll triggers writeback

s-cork:
address #307 

This was a bit of a nightmare
The docs say that user events are distinguished in a certain way
but they are not distinguished by the selectAll button clicks!

So we hack our way into signalling whether a selectAll was triggered by a user or not

steps taken:
- **HACK 1:** we add a click handler to the selectAll/deselectAll buttons that turns on the flag `_select_all_is_user`
- the selectAll button is clicked
- the changeAll function is called
- this raises the change event
- **Hack 2:** the changeAll function is wrapped so that it turns off the flag `_select_all_is_user`
  - we do this because the changeAll only triggers the change event if something has changed
- in the change event we also turn off the flag to avoid any writeback infinite loops

### DIFF
--- a/client_code/MultiSelectDropDown/__init__.py
+++ b/client_code/MultiSelectDropDown/__init__.py
@@ -7,6 +7,7 @@
 
 import anvil.js as _js
 from anvil import HtmlPanel as _HtmlPanel
+from anvil.js.window import Function as _Function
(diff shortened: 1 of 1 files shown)


meatballs:
Wow. What a PITA.

PR #321, opened 2022-04-21
Augment - mitigate against jquery changing the event type

s-cork:
documented here:
https://learn.jquery.com/events/event-extensions/


meatballs:
Can we have a changelog entry and then this is good to go?

Issue #329, opened 2022-05-01
Add width property to MultiSelectDropDown and make the default width fit the content of the placeholder text


Duncanr-glitch:
**Is your feature request related to a problem? Please describe.**
To make the dropdown's width by default fit the place holder's text 

**Describe the solution you'd like**
While achieving the above, it is best applied by adding that as the default width through an explicit width property on the multiselect dropdown

**Describe alternatives you've considered**
NA

**I am considering making a PR for this issue**
Yes, right away

s-cork:
see the reference docs:
https://developer.snapappointments.com/bootstrap-select/examples/#width

I think the default should probably be "auto"
and "fit" should be documented.

I think we can just use the API provided to us by the javascript library rather than writing our own.

PR #330, opened 2022-05-01
Added width property to MultiSelect Dropdown

Duncanr-glitch:


Duncanr-glitch:
This is to fix Issue #329

s-cork:
Hi @Duncanr-glitch thanks for the PR. 

You'll need to remove the unnecessary imports. 

As well as following the style guide. See the contributions page in the docs. 

If you pip install pre-commit and then `pre-commit run --all-files` it should fix the formatting. 

Also please take a look at the init method for this component. You already have the dom_node as an attribute so there's no need to redo that.

Once that's done I'll take a more in depth look at the rest of it.

Duncanr-glitch:
@s-cork Thanks for taking a look and giving good feedback for fixing my dumb mistakes. I should have fixed them all so that it's ready to be reviewed :)

meatballs:
Could we have an entry in the changelog for this?

client_code/MultiSelectDropDown/__init__.py
```diff
@@ -208,6 +216,29 @@ def visible(self, val):
...
+                _inner_dom_node.css("width", f"{value}px")
+            else:
+                self._dom_node.css("width", value)
+                _inner_dom_node.css("width", value)
+        else:
+            _inner_dom_node.css("width", "fit-content")
+            self._dom_node.css("width", "fit-content")
+
+        self._width = value
+
```
s-cork, line 241:
```suggestion
    width = _component_property("width", "data-width", lambda v: v or "auto")
```

note that this property won't be dynamic - but that's a separate issue for a separate pr

client_code/MultiSelectDropDown/form_template.yaml
```diff
@@ -11,6 +11,7 @@ properties:
 - {name: enable_select_all, type: boolean, default_value: false, group: interaction,
   important: false}
 - {name: visible, type: boolean, default_value: true, group: appearance, important: true}
+- {name: width, type: string, default_value: '', group: appearance}
```
s-cork, line 14: If we go with `"fit"` and `"auto"` as possible width values these should be documented

s-cork (changes requested):
I think this can be simplified.
The Bootstrap select component already has a width option we can tap into.
https://developer.snapappointments.com/bootstrap-select/examples/#width

client_code/MultiSelectDropDown/__init__.py
```diff
@@ -233,6 +264,10 @@ def _mk_popover(self, init_node, **event_args):
         # this is a bit of a hack - we're using the libraries private methods for this
         init_node(self._el.data("selectpicker")["$bsContainer"])
 
+    def form_show(self, **event_args):
+        """This method is called when the HTML panel is shown on the screen"""
+        self.width = self.width
+
```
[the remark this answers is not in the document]
Duncanr-glitch: This was based on my original implementation using just JQuery. Since the original property setter is before the dom node is actually present on the page, I just use a dummy property setting trigger in the form_show event. It doesn't have anything to do with the default, but I have added "fit-content" in _defaults because that's a good idea :)

client_code/MultiSelectDropDown/__init__.py
```diff
@@ -14,6 +14,14 @@
...
 from ._anvil_designer import MultiSelectDropDownTemplate
 
+# SPDX-License-Identifier: MIT
+#
+# Copyright (c) 2021 The Anvil Extras project team members listed at
+# https://github.com/anvilistas/anvil-extras/graphs/contributors
+#
+# This software is published at https://github.com/anvilistas/anvil-extras
+
+
```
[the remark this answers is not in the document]
Duncanr-glitch: Done in the next commit

client_code/MultiSelectDropDown/form_template.yaml
Duncanr-glitch: I have added fit-content in the next commit as the default_value. Do I need to write out a proper description, or is that enough documentation?

client_code/MultiSelectDropDown/__init__.py
Duncanr-glitch: This would be simpler, but if we can make it dynamic (which my current implementation does) so that when the width property is used it can be changed, why not? I very well might be missing another reason that we might prefer the simpler static option to the more verbose dynamic one, but I'm not sure what that is right now.

Duncanr-glitch:
@meatballs Added a line in the ChangeLog with the `width` property as a New Feature.

client_code/MultiSelectDropDown/__init__.py
s-cork: To make it dynamic is a one line change on my part. 

Certain properties like `enable_select_all` would also benefit from this change. It's a bug that I've not gotten around to fixing because no-one's complained. 

I'm happy to make that PR and for you to rebase on top of it. 

The Multiselect library we're using already has an api for defining the width so we should use it.

The short version is abstracted away and uses the library's api to implement options that the library already has an api for. I can't see a justification for not using it.

Duncanr-glitch: If its not too much trouble to you, I would like to make this dynamic -- seems like this is a property that's more prone to change than `enable_select_all` imho

s-cork: When you say dynamic are we talking about the same thing?

I mean you can set it in code after the component has been instantiated.

Duncanr-glitch: When I say dynamic, I mean that you can change the value through some piece of code after the component is first instantiated. With my OG implementation you can do that, but with the native component's property, I can't change the width after setting it initially.

s-cork:
@Duncanr-glitch where are we on this PR - the main branch now has support for dynamic properties for the MultiSelectDropDown component.

Some other thoughts:
- I'm not convinced that `fit` is the right default value. I'd prefer the default value to be as it currently is (i.e. empty) or `auto`
- I'll need to adjust how align works since an align value of `"full"` effectively makes the `width=100%`
  - no need to address that here - i can adjust that after this PR
- i'll also want to add the width as a dynamic property in the designer
- the docs should also add the `width` attribute
- I think a description would be useful in the yaml - this is visible as a tooltip when you hover over a property in the designer

PR #353, opened 2022-06-21
Retain existing roles when nav links are clicked

### LINKED ISSUE #352: Fix override of roles when navigation links are clicked
### Discussed in https://github.com/anvilistas/anvil-extras/discussions/345

<div type='discussions-op-text'>

<sup>Originally posted by **kushfest** June 10, 2022</sup>
Is there a clean way to have a headings in the navigation module? 

My navigation menu needs some organization and I'm hoping to add some headers. I was able to get it setup, but the headers don't have targets, so they give an error when clicked. Here's an example of what i've got, with CONTRACTS intended as a header:

image</div>

meatballs:
fixes #352

### DIFF
--- a/client_code/navigation.py
+++ b/client_code/navigation.py
@@ -60,16 +60,25 @@ def go_to(target):
     link.raise_event("click")
 
 
+def _reset_links():
+    for link in _links:
+        if isinstance(link.role, str):
+            link.role = [link.role]
+        try:
+            link.role = [r for r in link.role if r != "selected"]
+        except TypeError:
+            link.role = []
+
+
 def _default_link_click(**event_args):
     """A handler for navigation link click events
     * Clears the role of all links registered in this module
     * Set the calling link's role to 'selected'
     * Calls the relevant action for classic or hash routing
     """
-    for link in _links:
-        link.role = ""
+    _reset_links()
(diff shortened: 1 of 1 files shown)


client_code/navigation.py
```diff
@@ -60,16 +60,25 @@ def go_to(target):
...
 
 
+def _reset_links():
+    for link in _links:
+        if isinstance(link.role, str):
+            link.role = [link.role]
+        try:
+            link.role = [r for r in link.role if r != "selected"]
+        except TypeError:
+            link.role = []
```
s-cork, line 70: What TypeError are we catching here?

s-cork (approved):
LGTM - one question

client_code/navigation.py
meatballs: Probably OTT but I wanted to handle any possibility that role might be None and therefore not iterable.

meatballs: Happy to get rid of it if you prefer

meatballs: Actually, just testing it and None seems to appear sometimes, so I've put this back in

meatballs: Ha! Yes, because of kwargs, as you say!

PR #355, opened 2022-06-21
Fix bug where existing roles get removed on click

meatballs:


client_code/navigation.py
```diff
@@ -78,7 +78,7 @@ def _default_link_click(**event_args):
     """
     _reset_links()
     link = event_args["sender"]
-    link.role = ["selected"]
+    link.role.append("selected")
```
s-cork, line 81: This won't trigger a change to the dom, you'll need to assign back to the role property

PR #381, opened 2022-11-02
Check the window before dynamically importing scripts

s-cork:
This will allow a service to import these scripts in advance 👍


meatballs:
Looks good. Can we just have a changelog entry?

PR #384, opened 2022-11-11
uuid - add deprecation warning

s-cork:


s-cork:
@meatballs - need to run the version bump across the codebase I think

meatballs:
I think it's just that the version is missing from deprecate.py

s-cork:
oh my bad! One day i'll add that to the pre-commit script somehow

PR #393, opened 2022-12-15
Add args and kwargs to navigation.open_form

meatballs:


client_code/navigation.py
```diff
@@ -47,9 +47,9 @@ def set_title(text):
     _title_label.text = text
 
 
-def open_form(form_name, full_width=False):
+def open_form(form_name, full_width=False, *args, **kwargs):
```
s-cork, line 50: Would it be breaking to make full_width a keyword only argument? If a user did want to make use of args they'd probably get hit with the first arg being used for full_width.

s-cork (approved):
One comment.

PR #400, opened 2023-02-28
Add persistence module, docs and tests

meatballs:


meatballs:
Differences here from the work in anvil-labs:

* `PersistedClass` has a `key` class attribute which must be defined on any subclass
* `PersistedClass.get` is now a classmethod
* `PersistedClass` has a `_cache` class attribute which is used by `get` and populated by `get` and `search`
* `PersistedClass.search` has a `lazy` parameter. Its default value is `False` whereby the method returns a list of instances and caches each of those. When `True`, it returns a generator and caches nothing.

These changes are mostly to enhance the use of the persistence module with hash routing. It allows us to use simple keys in a route's url without having to make unnecessary server calls to fetch objects that have already been instantiated.

Bonus - This module, hash routing and tabulator work very nicely together!!

client_code/persistence.py
```diff
@@ -0,0 +1,166 @@
...
+                is_persisted_class = issubclass(value, PersistedClass)
+            except TypeError:
+                is_persisted_class = False
+
+            if is_persisted_class:
+                setattr(cls, attr, LinkedClass(cls=value, linked_column=attr))
+
+    @classmethod
+    def search(cls, lazy=False, *args, **kwargs):
+        rows = anvil.server.call(f"search_{cls.__name__.lower()}", *args, **kwargs)
```
s-cork, line 101:
```suggestion
        rows = anvil.server.call(f"search_{cls._snake_name}", *args, **kwargs)
```

client_code/persistence.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __init__(self, store=None, *args, **kwargs):
+        self._store = store or {}
+        self._delta = {}
+
+    def __getattr__(self, key):
+        if self._delta and key in self._delta:
+            return self._delta[key]
+
+        return dict(self._store).get(key, None)
```
s-cork, line 129: this call to `dict` seems unnecessary and we seem to be losing any AttributeErrors for missing keys
if you decide to use `q.fetch_only` in your `get` then you've just negated that here

client_code/persistence.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __getitem__(self, key):
+        return getattr(self, key)
+
+    def __setattr__(self, key, value):
+        is_private = key.startswith("_")
```
s-cork, line 135: is this documented?

client_code/persistence.py
```diff
@@ -0,0 +1,166 @@
...
+        is_descriptor = hasattr(self.__class__, key) and hasattr(
+            getattr(self.__class__, key), "__set__"
+        )
+        if is_private or is_descriptor:
+            object.__setattr__(self, key, value)
+        else:
+            self._delta[key] = value
```
s-cork, line 142: This seems to mean we're not optimistic with our updates?
i.e. you don't get the `_delta` until you've called `update()`

```python
book.foo = "bar"
book.foo # None
book.update()
book.foo # "bar"
```

s-cork (approved):

client_code/persistence.py
meatballs: Ugh! How did that error creep back in?!!!

client_code/persistence.py
meatballs: Yep, that's a throw back to a previous attempt which I'd forgotten all about. I'll get rid...

client_code/persistence.py
meatballs: Nope!

meatballs: It is now

client_code/persistence.py
s-cork: missed the line of code in `__getattr__` that did this

client_code/persistence.py
```diff
@@ -126,7 +126,7 @@ def __getattr__(self, key):
         if self._delta and key in self._delta:
             return self._delta[key]
 
-        return dict(self._store).get(key, None)
+        return self._store.get(key, None)
```
s-cork, line 129: but if a `row` is a store we don't have a `.get()`?

meatballs: shite. That's exactly what it was there for. Must slow down... Must slow down...

s-cork:
```suggestion
        try:
            self._store[key]
        except (KeyError, anvil.tables.TableError):
            raise AttributeError(key)
```

then adjust `__getitem__`?

meatballs: yeah, I like that. Much nicer. Why would `__getitem__` need to change?

s-cork: To throw a KeyError

meatballs: Right - just so that attribute access raises an AttributeError but dictlike access raises a KeyError?

s-cork: Yeah. That was my thinking.

meatballs: Gotcha. I couldn't work out what logic needed to change!

client_code/persistence.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+
+    def __set__(self, instance, value):
+        raise AttributeError(
+            "Linked Class instance is already set and cannot be changed"
+        )
+
+
+class PersistedClass:
+    _cache = {}
```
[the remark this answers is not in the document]
meatballs: Sounds wise. On it...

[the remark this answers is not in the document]
meatballs: It's also a little cleaner for it to live alongside `_snake_name`

meatballs:
Merge if/when you're happy with this and I'll do a release

client_code/persistence.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+        if is_private or is_descriptor:
+            object.__setattr__(self, key, value)
+        else:
+            self._delta[key] = value
+
+    def __setitem__(self, key, value):
+        setattr(self, key, value)
+
+    def add(self, *args, **kwargs):
```
s-cork, line 154: I think it would be nice to document this. The api isn't obvious since it's an instance method so you need to create the object before adding it, and the server call should return the store.

meatballs: Looking at the docs, I think a bit of rewrite might not be a bad thing...

meatballs:
Thank you! Release incoming...

PR #405, opened 2023-03-17
Pre commit bump version support

s-cork:
Our pre-commit hooks don't fix versioning 
this commit adds that support
a lot of back and forth with chat gpt-4

But I think it's there


s-cork:
to test locally - `pre-commit install`
then change a version number in a python file
stage those and try to commit them
and it should auto fix

PR #432, opened 2023-07-15
trigger('writeback'): deprecate and suggest alternative

### LINKED ISSUE #429: Deprecate trigger writeback
This is now possible to do directly from anvil so no need for our implementation, which knows too much about the anvil internals

I don't think the native version is yet documented but we can just deprecate it with a warning and instructions for how to do it natively

s-cork:
close #429


s-cork:
Already added as part of #488

PR #437, opened 2023-08-03
routing: fix alerts no longer block

s-cork:
This PR fixes alerts 
The previous implementation relied on knowing that alerts were bootstrap alerts
This is no longer the case
Instead it provides a `routing.alert`, which wraps `anvil.alert`

The behaviour is as before, dismissible alerts with close, non-dismissible alerts will block

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# Unreleased
+
+## Bug Fixes
+* routing was no longer dismissing alerts on navigation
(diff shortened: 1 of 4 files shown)


meatballs:
LGTM. Could we just have a changelog entry and I reckon that's good to go. Probably justifies a release too.

s-cork:
Yeah agreed - probably warrants a release.

PR #442, opened 2023-08-11
Lazy route

s-cork:
tagging @steininge and @jshaffstall  who might be interested in this one following this forum post:

https://anvil.works/forum/t/lazy-loading-for-forms/18166/2

```python

@routing.lazy_route('article', url_keys=['id', routing.ANY], title="Article-{id} | RoutingExample")
def article_route():
    from ..ArticleForm import ArticleForm
    return ArticleForm
 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## New Features
+* `routing.lazy_route` - allows you to lazily load Forms whilst using routing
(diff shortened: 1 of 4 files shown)


s-cork:
Yeah exactly. That would be an absolutely valid pattern.

steininge:
I like it. Thank you, @s-cork!

meatballs (approved):

Issue #434, opened 2023-07-28
Client Side Test Runner Component / Dashboard


yahiakala:
**Is your feature request related to a problem? Please describe.**
I would like a visual interface to run client side tests, either in a general way (run all tests) or granular way (run specific tests). **Would this make sense for the anvil-extras project?**

**Describe the solution you'd like**
- I have created a custom form with nested repeatingpanels that, with a few inputs, will automatically display all of the test modules, test classes, and test methods in the client code.
- Buttons will appear that allow you to run tests in the anvil app either on the method level, testclass level, module level (or RUN ALL tests)
- Personally, I have it set up to run all my server tests by calling them from the client side
- I have a top menu button that opens the form, that only appears when I am in my debug environment

**Describe alternatives you've considered**
- I have considered other ways to organize my test code according to some popular posts on the Anvil Forum. This is the one I have referred to the most: https://anvil.works/forum/t/wiki-best-practices-test-driven-development-with-anvil/10652
- However, I want to be able to test code within my anvil app's debug environment and test UI components
- I also don't want to use a specific design pattern (removing all logic and server calls from forms)

**Additional context**
<img width="1046" alt="image" src="https://github.com/anvilistas/anvil-extras/assets/24422258/94405199-3518-4c5c-9bae-1b8649c6b747">

Here's how I'd load the form:
<img width="618" alt="image" src="https://github.com/anvilistas/anvil-extras/assets/24422258/fd5a7071-d724-4648-bdb9-55ec62dfe313">

Here's how I'd show the top menu button (default hidden) if in debug mode:
<img width="618" alt="image" src="https://github.com/anvilistas/anvil-extras/assets/24422258/1def3204-c401-4537-a187-565b283e1ebb">

**I am considering making a PR for this issue**
Yes. I am still reading through the contribution guidelines and will probably have to cross some T's and dot some i's, but it is pretty much ready if you're interested.

PR #443, opened 2023-08-15
Add Unit Testing Component

yahiakala:
I have created a unit testing component for running client side tests in a UI. I have also added a test module and some documentation.

This is a follow up from this issue: https://github.com/anvilistas/anvil-extras/issues/434

I went through the contributing guide and made sure to run pre-commit as well as the sphinx build to make sure the docs were built correctly.

Here's how the unit testing component looks with no roles:

image


s-cork (commented):
some initial thoughts:

- the templates seem quite similar - could we refactor some of the common parts
- unittests have `setUp()` and `tearDown()` that you should be calling 
- this doesn't work with `self.assertEqual()` and other unittest methods - which seems problematic to me - possibly this is Skulpt issue

I'd probably want to put this in anvil labs first
Iterate on it
See if it has appeal/traction before moving it to anvil-extras

yahiakala:
- Any ideas on how to refactor the common parts? I too want to cut down on repetitive code but wasn't sure how. There are some important differences between the nested forms but the templates are pretty much identical.
- I can add the calls to setUp and tearDown if those methods exist. Currently, I am running the test methods in a loop so I'll have to add those calls explicitly.
- I just tested self.assertEqual() and it works for me! In addition to the following:

```python
self.assertEqual(True, True)
self.assertNotEqual(True, False)
self.assertTrue(True)
self.assertFalse(False)
self.assertIsNone(None)
self.assertIsNotNone(4)
self.assertIn(3, [1, 2, 3, 4, 5])
self.assertNotIn(10, [1, 2, 3])
```

I'm happy to move this over to anvil-labs to continue the discussion.

s-cork:
did you check whether it failed correctly? `self.assertEqual(1, 2)`

If it's a `unittest.TestCase`, then `setUp` and `tearDown` methods are guaranteed to exist
They are implemented as noop methods in the base class
https://github.com/python/cpython/blob/02538200b34c7f51852b8759f78c31da6d726964/Lib/unittest.py#L127-L133

For refactoring - either consider a super class that all the Template classes could inherit from
Or consider each Template can add a common component that does the common things
It really depends, but once you've written the same thing 3 times then you know it's time to refactor.

Yes - I would vote for moving this over to anvil-labs for further discussion and iteration

PR #460, opened 2023-09-17
navigation - support a condition

### LINKED ISSUE #383: Make Navigation module aware of Authorisation module
It would be nice if the Navigation object could be made aware of the permissions that a user has. Eg show/hide items depending on the perms. 

As an alternative it would be helpful if the navigiation could dynamically hide/show items. Currently only Anvil events are supported for this. A callable to determine visibility would be more flexible.

s-cork:
close #383

naming suggestions welcome


meatballs:
Nice!

meatballs (approved):

PR #465, opened 2023-09-24
View transitions

s-cork:
This PR adds transitions to page navigation for the routing and navigation module

**To decide**:
Should this be opt in or opt out?
Presumably the next release will be a major version bump so we can decide now
This PR does opt in

See this site for the transitions in action
They're pretty subtle

https://dev-routing-example.anvil.app


meatballs:
Nice. I think opt out is the better way to go.

PR #467, opened 2023-09-29
Tabs resize observer

s-cork:
Essentially if the Tabs change size the indicator should also move to reflect this
At the moment the indicator stays fixed which is odd


meatballs:
Just the changelog and this is good to go

PR #468, opened 2023-09-29
Pivot: Don't completely fail if items is not set

s-cork:
Useful if this was instantiated in the designer and self.items was None


meatballs:
Nice! Can we have a changelog entry?

PR #469, opened 2023-09-29
switch: be more like a CheckBox in the designer

s-cork:
This is an interesting PR
In that it will break the Switch element in the Designer (for now 😉)


s-cork:
(draft for now)

PR #470, opened 2023-09-29
Switch: make it work with undocumented designer api

s-cork:
doesn't change anything from the outside


s-cork:
This one I think needs a slight tweak before merging - converting to draft for now

PR #483, opened 2023-11-08
improve get_rgb method to account for alpha values

s-cork:
@telemix probably wants to fade the active_background of a tab
So it would be better to use a hex color with an alpha value e.g. `#2196F344` where 44 is the alpha part

This PR supports that


meatballs:
Just a changelog entry needed for this one

PR #485, opened 2023-11-13
Retry a call to the store if it fails

s-cork:
This fixes a bug in IOS where a web worker is sleeping during an indexed db transaction
and then waking up and the transaction falls over
retrying the transaction fixes the bug

I wrote a PR direct to local forage library to fix this
They already do retries but don't retry on this particularly error type
If that gets merged we can probably remove this code and just update our version of local forage

@rhurlbatt  this is the fix for #484

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 - Tabs - adds active_background property
   https://github.com/anvilistas/anvil-extras/issues/481
 
+## Bug Fixes
+- storage - fix a bug in ios when indexed db accessed in a webworker and closed
+  https://github.com/anvilistas/anvil-extras/discussions/484
+
 # v2.5.2 25-Oct-2023
 * storage - fix bug with deserializing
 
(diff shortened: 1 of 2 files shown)


s-cork:
Current dependencies on/for this PR:
* main
  * **PR #485** <a href="https://app.graphite.dev/github/pr/anvilistas/anvil-extras/485?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/anvilistas/anvil-extras/485?utm_source=stack-comment).

rhurlbatt:
Thank you Stu. So the idea is that if it fails because it is asleep a single retry is enough to mitigate the error?
Looks like the PR to localForage is a !huge change :)

s-cork:
Yeah. That's the idea. With local testing just wrapping store access in a single try except was enough. This PR just makes sure we only wrap store access in retry logic.

PR #492, opened 2023-11-29
fix: cant remove handlers after wrapping in augment

### LINKED ISSUE #490: Issue with latest Anvil Extra changes
Hello, I am encountering a bug I believe to be related to the pull request #488 that occurred a few days prior. 

When I use a previously saved version of anvil-extras in my project, everything works as it should. However, when I clone the latest version I get an error & I am unable to load any content inside of the content panels.

Here is the traceback:
Screenshot from 2023-11-28 11-40-52

I noticed #488 is related to the augments.py module, the module seemingly causing the error I am currently having.

**Describe the bug**
The content panel is unable to load any content.

**Version**
Using the latest version from the Github repository, I encounter this bug. 
(I need my customers to be able to run a shell script that runs the command 'git clone https://github.com/anvilistas/anvil-extras')

**Screenshots**
Here is a screenshot of what my anvil project looks like when I encounter the bug (notice the empty content panel.)
image

I can click on the left-hand side nav bar but each form looks the same, nothing is able to load.

**Additional context**
If there is a way for me to run 'git clone' but use a previous build of anvil extras (before pull request #488) that would also be a solution I can try out.

Thanks for the help. Please let me know if there is any other information I can provide to help

s-cork:
close #490

shouldn't need a change log since this is unreleased
does the fast thing if we're don't need to wrap an event handler for augment
keeps a cache of augmented handlers so that removing handlers works

### DIFF
--- a/client_code/augment.py
+++ b/client_code/augment.py
@@ -134,23 +134,29 @@ def _add_event(self, event_name):
         self.remove_event_handler(event_name, _noop)
 
 
+@_cache
+def _get_handler(fn, event):
+    @_wraps(fn)
+    def wrap_handler(*args, **kws):
+        kws["event_name"] = event
+        return fn(*args, **kws)
+
+    return wrap_handler
+
+
 def wrap_event_method(method):
     old_method = getattr(_Component, method)
 
     @_wraps(old_method)
     def wrapped(self, event, *args, **kws):
         key = (type(self), event)
-        remapped = _prefix + event if key in _remap else event
+        if key not in _remap:
+            return old_method(self, event, *args, **kws)
+
+        remapped = _prefix + event
 
         if len(args) == 1 and callable(args[0]):
-            fn = args[0]
-
-            @_wraps(fn)
-            def wrap_handler(*args, **kws):
-                kws["event_name"] = event
-                return fn(*args, **kws)
-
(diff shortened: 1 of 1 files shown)


s-cork:
Current dependencies on/for this PR:
* main
  * **PR #492** <a href="https://app.graphite.dev/github/pr/anvilistas/anvil-extras/492?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/anvilistas/anvil-extras/492?utm_source=stack-comment).

s-cork:
i think this is probably ok, but do we want to implement a maxsize lru_cache instead?
i can easily see that the cache could get quite large if augment handlers are being used in a datagrid row template

rhurlbatt:
For what it's worth, we do use augment in data rows to hide and show action
buttons dynamically so anything to optimise that would be greatly
appreciated.

s-cork:
Noted. 

If you're really looking to optimise you might want to consider attaching event listeners directly to dom nodes with anvil.js.get_dom_node. 

Augment is really for convenience these days since it was written before anvil.js existed.

PR #501, opened 2024-02-07
routing: fix check stale

s-cork:
fixes #500

tag @rhurlbatt


meatballs:
Just a changelog entry and I reckon this is good to go

PR #517, opened 2024-02-29
Fix lever color in switch component

yahiakala:
The switch component wasn't incorporating the transparency correctly. As a result, the bar behind the switch was turning white when checked.

I just changed one line:

    background-color: rgba(var(--color), .5);

To:

    background-color: rgba(var(--color) / .5);
    
    
All the other changes (to the yaml files) were done by the Anvil Editor.


s-cork (approved):
Lgtm

s-cork:
Oops forgot to request a change log item. Could you add one @yahiakala and pr that

PR #546, opened 2024-07-17
Persistence data descriptor

uglyfruitcake:
The persistence module expressly prohibited changing linked class instances. It instead raised an error in the linked class descriptor. 
This PR replaces that behaviour by recording the change in the parent object's _delta.


meatballs:
I've had a poke around at this whilst it was being worked on, so I'm happy with it. I'll let @s-cork review it before we merge.

PR #551, opened 2024-08-01
Add colors module

meatballs:
A colors.py module with a Manager class to handle switching of color theme via code.


s-cork:
Some thoughts:
A variant of the M3 theme is probably the most common. But that will probably change if/when something replaces M3.

I think apps probably have a dark and light theme but wont have different named colour schemes they are concerned with. This makes the common thing to do (set theme dark/light) a bit more verbose.

Maybe consider `set_theme`
And `set_variant`

And maybe set_theme can take a string or dictionary. 

```
my_theme = {"light": ...}

manager.set_theme(my_theme, variant="light")
manager.set_variant("dark")

```

The current api also suggests we might have multiple variants other than light and dark. 

Do we want that?

Could just use `set_dark(True)` or `set_theme(my_theme, dark=True)`


Beyond the scope of this PR (probably):
If we only think about dark and light themes should a colour manager be responsible for "auto" as an option? ie use the browser media query to determine whether the user prefers light or dark mode. 

And possibly an api for supporting remembering what the preference is in local storage. 
And then we get into the realm of avoiding a flicker of light mode if the user actually prefers dark mode.

meatballs:
> A variant of the M3 theme is probably the most common. But that will probably change if/when something replaces M3.

Agreed. That's why I went with a class that can take a dict of colours in its init - so that those can be anything at all. If/when something new comes along, we could always add more constants to provide those and probably even remove M3 as the default.

> I think apps probably have a dark and light theme but wont have different named colour schemes they are concerned with. This makes the common thing to do (set theme dark/light) a bit more verbose.

I've had a couple of instances where a multi-tenant app needed to swap theme entirely depending on the tenant of the logged in  user. However, I agree that just swapping between light and dark is more common so I should add something to make that simple.

> Beyond the scope of this PR (probably):
> If we only think about dark and light themes should a colour manager be responsible for "auto" as an option? ie use the browser media query to determine whether the user prefers light or dark mode.
>
> And possibly an api for supporting remembering what the preference is in local storage.
> And then we get into the realm of avoiding a flicker of light mode if the user actually prefers dark mode.

I did wonder about doing some of that but thought it could make a nice further addition at some point in the future.

docs/guides/modules/theme.rst
```diff
@@ -0,0 +1,99 @@
...
+
+    color_manager.set_scheme("Material", "dark")
+
+You can define your own color schemes and use them in the same way. For a material
+design scheme, follow the `Creating a custom Material Design 3 colour scheme <https://anvil.works/docs/how-to/creating-material-3-colour-scheme>`_ docs.
+
+.. code-block:: python
+
+   my_schemes = {
+        scheme_one: {
```
s-cork, line 61:
```suggestion
        "scheme_one": {
```

client_code/theme.py
```diff
@@ -0,0 +1,306 @@
...
+            "Dark Overlay 2": "rgba(44, 22, 13, 0.12)",
+            "Primary Overlay 1": "rgba(154, 69, 35, 0.05)",
+            "Primary Overlay 2": "rgba(154, 69, 35, 0.08)",
+            "Primary Overlay 3": "rgba(154, 69, 35, 0.11)",
+        },
+    },
+}
+
+
+class Colours:
```
s-cork, line 272: Consider `Colors` over `Colours`

meatballs:
I've refactored - aiming to cope with as wide a range of possible schemes whilst making it convenient to do the most frequent stuff.

docs/guides/modules/theme.rst
```diff
@@ -0,0 +1,99 @@
...
+
+If you want to set both the scheme and variant at the same time, you can use the `set_scheme` method:
+
+.. code-block:: python
+
+    colours.set_scheme("Material", "dark")
+
+.. code-block:: python
+
+    color_manager.set_scheme("Material", "dark")
```
s-cork, line 53:
```suggestion
    colours.set_scheme("Material", "dark")
```

docs/guides/modules/theme.rst
```diff
@@ -0,0 +1,99 @@
...
+
+You can define your own color schemes and use them in the same way. For a material
```
s-cork, line 55: spelling of color to match rest of docs

client_code/theme.py
meatballs: Yeah, I've been torn over this one. In the end, I went with the same spelling as is used in the Anvil IDE - the british one.

client_code/theme.py
```diff
@@ -0,0 +1,306 @@
...
+    def __init__(self, schemes=None, default_scheme=None, default_variant=None):
```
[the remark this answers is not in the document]
meatballs: ah. that's a throwback to before scheme and variant were properties. I'll sort that...

client_code/theme.py
s-cork: typically anvil does british spelling for docs, and things in the designer, but US spelling for code names

e.g. `anvil.app.theme_colors`
and in the yaml
`parameters.yaml -> color_scheme -> colors`

I don't mind either way
personally I always default to choosing US spelling for code names

client_code/theme.py
```diff
@@ -0,0 +1,306 @@
...
+        self._variant = default_variant or tuple(self.schemes[self._scheme].keys())[0]
+        self._set_scheme(self.scheme, self.variant)
+
+    def _set_scheme(self, scheme, variant):
+        try:
+            anvil.app.theme_colors.update(self.schemes[scheme][variant])
+        except KeyError:
+            raise ValueError(f"{scheme} {variant} is not defined.")
+
+    def set_scheme(self, scheme, variant):
```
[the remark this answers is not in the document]
meatballs: If you just want to change the scheme, you could set the `scheme` property. This method was intended to be for changing both at once.

docs/guides/modules/theme.rst
meatballs: oops!

client_code/theme.py
meatballs: ah yes. I'd forgotten that little wrinkle! I'll swap the lot to the US version. Even I'm used to typing that in code!

s-cork (approved):

PR #572, opened 2024-11-01
Added support for Logger usage in messaging.py

### LINKED ISSUE #569: Use a custom format logger to print Publisher messages
**Is your feature request related to a problem? Please describe.**
When using the Publisher class to publish or subscribe to a channel you can pass a parameter called "with_logging" to override the default way for printing in the console output.

Printing, however, would break the flow of logs when using a custom format of Logger (also from anvil-extras).

It would be interesting if the Publisher construtor and it's methods could receive a Logger instance to use to print in the correct format.

image

image

image

The Level could be INFO (by default or as parameter) and the same message printed now could be the msg parameter of the logger format.

GDuroS:
Added support for receiving a Logger instance as the with_logging parameter in Publisher methods and constructor.
Added support for defining the level for logging in Publisher methods (defaults to INFO).

close #569

### DIFF
--- a/client_code/messaging.py
+++ b/client_code/messaging.py
@@ -6,6 +6,7 @@
 # This software is published at https://github.com/anvilistas/anvil-extras
 
 from logging import Logger
+
 __version__ = "3.0.0"
 
 
@@ -41,7 +42,7 @@ class Publisher:
                 with_logging.log(
                     log_level,
                     f"Published '{message.title}' message on '{channel}' channel to "
-                    f"{len(subscribers)} subscriber(s)"
+                    f"{len(subscribers)} subscriber(s)",
                 )
             else:
                 print(
@@ -49,7 +50,9 @@ class Publisher:
                     f"{len(subscribers)} subscriber(s)"
                 )
 
-    def subscribe(self, channel, subscriber, handler, with_logging=None, log_level=None):
+    def subscribe(
+        self, channel, subscriber, handler, with_logging=None, log_level=None
+    ):
         if with_logging is None:
             with_logging = self.with_logging
         if log_level is None:
@@ -74,7 +77,9 @@ class Publisher:
             ]
         if with_logging:
             if isinstance(with_logging, Logger):
-                with_logging.log(log_level, f"Removed subscriber from {channel} channel")
+                with_logging.log(
+                    log_level, f"Removed subscriber from {channel} channel"
+                )
             else:
                 print(f"Removed subscriber from {channel} channel")
 
@@ -87,6 +92,8 @@ class Publisher:
         del self.subscribers[channel]
         if with_logging:
             if isinstance(with_logging, Logger):
-                with_logging.log(log_level, f"{channel} closed ({subscribers_count} subscribers)")
+                with_logging.log(
+                    log_level, f"{channel} closed ({subscribers_count} subscribers)"
+                )
             else:
                 print(f"{channel} closed ({subscribers_count} subscribers)")
--- a/client_code/messaging.py
+++ b/client_code/messaging.py
@@ -5,7 +5,7 @@
 #
 # This software is published at https://github.com/anvilistas/anvil-extras
 
-from logging import Logger as _Logger
+from logging import Logger as _Logger, INFO
 
 from .utils._warnings import warn as _warn
 
@@ -25,10 +25,11 @@ class Subscriber:
 
 
 class Publisher:
+    default_logger = None
+    default_log_level = INFO
+
     def __init__(self, logger: _Logger = None, **kwargs):
-        if logger is None:
-            logger = _Logger(name="publisher", format="{msg}")
-        self.logger = logger
+        self.logger = logger or self.default_logger
         self.subscribers = {}
         self._deprecation_warnings(**kwargs)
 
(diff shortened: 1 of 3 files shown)


s-cork (commented):
I think combining these lots into a private log method would be handy, since the logic is mostly the same except for the msg.

meatballs:
Injecting a logger is a much better solution than the original implementation.


However, if we're moving to that, I prefer to make a breaking change and do it cleanly.

Rather than keeping the old param name and making it handle both the old and new behaviour, we should drop it, add a new `logger` param with default of None and instatiate a default  logger instance if nothing is provided.

GDuroS:
@meatballs Understood and agreed. This will, however, as sou said it, be a breaking change. What about people who are passing values to `with_logging`, or that do not want to log anything? Shouldn't we maintain compatibility?

What I thought was:

- Publisher constructor receives a Logger instance of None, instantiating a new one if None.
- Methods always use the same logger instance from the Publisher. They can, however, inform the level or None if they don't want to log.

What do you think? This will still break compatibility. What should we do? Do it like this and warn users, accept the `with_logging` param as a kwarg but ignore it? What is the best approach?

s-cork:
If we go down a breaking change route I would personally accept the with_logging kwarg for now, and print a warning if it's used and that it will be removed in a future version. 

See `_warnings.py` for a utility function you can use.

GDuroS:
Ok, I did some changes. I did like I said in my previous comment, receiving the Logger instance at creation of the publisher and having a private method to handle the warning messages for deprecation. I dind't ask for a log level at each method, instead using INFO as default, since you can disable the messages simply by using a custom logger with higher level.

Let me know if you think I should change anything else.

client_code/messaging.py
```diff
@@ -20,46 +25,49 @@ def __init__(self, subscriber, handler):
 
 
 class Publisher:
-    def __init__(self, with_logging=True):
-        self.with_logging = with_logging
+    def __init__(self, logger: _Logger = None, **kwargs):
+        if logger is None:
+            logger = _Logger(name="publisher", format="{msg}")
+        self.logger = logger
```
s-cork, line 31:
```python

class Publisher:
    default_logger = Logger(name="publisher", format="{msg}")
    def __init__(self, logger: _Logger = None, **kwargs):
        self.logger = logger or self.default_logger
```

This way we can also change the log level of the default logger
like 

```python
from anvil_extras.messaging import Publisher
from anvil_extras.logging import ERROR

Publisher.default_logger.log_level = ERROR
# or
Publisher.default_logger = Logger(...)
```

But I think this api suffers from not having an easy way to turn off logging.

GDuroS: Yeah, I agree, but every solution that I think goes back to having the `with_logging` parameter or having an hybrid approach like the first one that I did.

Also, I changed the docs to show info about the usage of loggers in the messaging module.

Will fix what you commented right now.

s-cork: yeah curious to hear @meatballs thoughts on this api

GDuroS: For disabling the log, what about returing the `with_logging` option to each method, so that you can disable for specific calls. 

We could also add `with_logging` to the constructor and set logger to None if disabled. 

Additionally, we could create a `classmethod` to disable the `default_logger`, setting it to None so that you could disable it (or change it) globally, like in other modules.

Then, before each logging action, we check if `self.logger` is None (or with_logging is True).

s-cork: Yeah. And actually most of the time you probably want logging off by default. 

If you could add updating the docs to this PR that'd be great.

meatballs: The default logger could be a null logger. i.e. if you pass no logger, you get no logging.

GDuroS:
- Changed the code to use a default_logger as None. If no logger is passed at instance creation, the default logger will be used.
- default_logger can be changed globally using @s-cork 's suggestion. Log level can also be changed the same way.
- If no logger is set (default was not changed and no logger was passed at creation), no message will the printed (as was @meatballs ' suggestion).
- log will be made using global default_level (that can also be changed ).
- Docs were updated again with those informations

s-cork (approved):
lgtm

tests/test_publisher.py
```diff
@@ -2,7 +2,9 @@
 
 
 def test_default_logging(capsys):
-    publisher = messaging.Publisher()
+    from client_code.logging import DEBUG, Logger
+
+    publisher = messaging.Publisher(logger=Logger(level=DEBUG))
```
s-cork, line 7:
```suggestion
    publisher = messaging.Publisher(logger=Logger(level=DEBUG, format="{msg}"))
```

meatballs:
Rather than the default logger being None, I had in mind that it would  be a logger instance that logs to /dev/null

That way, we don't need the condition to check if it's None every time we log a message.

s-cork:
@meatballs - the logging implementation doesn't have a null logger

would you suggest creating one?

```python
# logging

class _NullStream:
    def write(self, *args, **kws): pass
    def flush(self, *args, **kws): pass

# Create a logger that writes to the null stream
null_logger = Logger(name="null_logger", stream=_NullStream())
```

meatballs:
Ah. I should probably have checked that first!

In that case, let's merge thus as is. We can always come back and refactor with this addition to the logging module. It wouldn't affect the api here.

s-cork:
@GDuroS - looks like we forgot to add a note to CHANGELOG.md - could you add a PR with a change log entry

PR #574, opened 2024-11-09
Add default null logger to messaging

meatballs:


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 ## Enhancements
 * tabs - tab color can now be a css var and better support for css colors in general
   https://github.com/anvilistas/anvil-extras/pull/570
+* messaging - support custom loggers in messaging module
+  https://github.com/anvilistas/anvil-extras/issues/569
```
s-cork, line 14:
```suggestion

## Deprecations
- messaging - with_logging argument is deprecated and maybe removed in a future version
  logging messages is now off by default - to turn it on provide a custom `logger` argument
  https://github.com/anvilistas/anvil-extras/pull/572/files

```

while we're here - add missing changelog entry

client_code/messaging.py
```diff
@@ -26,11 +29,10 @@ def __init__(self, subscriber, handler):
 
 
 class Publisher:
-    default_logger = None
     default_log_level = INFO
 
     def __init__(self, logger: _Logger = None, **kwargs):
```
s-cork, line 34:
```suggestion
    def __init__(self, *, logger: _Logger = None, **kwargs):
```

Make logger keyword only (not completed relevant to this PR but while we're here)

s-cork (approved):
lgtm - a couple of suggestions while we're touching messaging

PR #575, opened 2024-11-12
Add serialisation of persistence delta

meatballs:


s-cork (approved):

client_code/persistence.py
```diff
@@ -162,15 +162,22 @@ def __setattr__(self, key, value):
     def __setitem__(self, key, value):
         setattr(self, key, value)
 
+    def __eq__(self, other):
+        return self._store == other._store
```
s-cork, line 166:
```suggestion
        if not instance(other, type(self)):
            return NotImplemented
        return self._store is not None and self._store == other._store
```

s-cork: oops sorry for the typo

client_code/persistence.py
```diff
@@ -162,15 +162,24 @@ def __setattr__(self, key, value):
     def __setitem__(self, key, value):
         setattr(self, key, value)
 
+    def __eq__(self, other):
+        if not instance(other, type(self)):
```
s-cork, line 166:
```suggestion
        if not isinstance(other, type(self)):
```

PR #588, opened 2024-12-21
Added height and width properties to both progress bars

### LINKED ISSUE #428: Set the height of a progress bar
**Is your feature request related to a problem? Please describe.**
It seems when using your progress bar one has to stick to this thin size.
This makes it hard to integrate it nicely into our page design.

**Describe the solution you'd like**
It would be great to have an option to set the height of the progress bar.

I would, but don't know where to start :(

Duncanr-glitch:
Adds the feature in issue close  #428:

Added height for both determinate and indeterminate progress bars.

### DIFF
--- a/client_code/ProgressBar/Determinate/__init__.py
+++ b/client_code/ProgressBar/Determinate/__init__.py
@@ -42,7 +42,7 @@ class Determinate(DeterminateTemplate):
     @height.setter
     def height(self, value):
         self._height = value
-        self.indicator_dom_node.style.setProperty("height", _css_length(value))
+        self.indicator_dom_node.style.setProperty("height", _css_length(value or "3px"))
 
     @property
     def width(self):
@@ -51,7 +51,7 @@ class Determinate(DeterminateTemplate):
     @width.setter
     def width(self, value):
         self._width = value
-        self.dom_node.style.setProperty("width", _css_length(value))
+        self.dom_node.style.setProperty("width", _css_length(value or "3px"))
 
     @property
     def track_colour(self):
--- a/client_code/ProgressBar/Determinate/__init__.py
+++ b/client_code/ProgressBar/Determinate/__init__.py
@@ -37,21 +37,12 @@ class Determinate(DeterminateTemplate):
 
     @property
     def height(self):
-        return self._height
+        return self._props.get("height")
 
     @height.setter
     def height(self, value):
-        self._height = value
-        self.indicator_dom_node.style.setProperty("height", _css_length(value or "3px"))
-
-    @property
-    def width(self):
-        return self._width
-
-    @width.setter
-    def width(self, value):
-        self._width = value
-        self.dom_node.style.setProperty("width", _css_length(value or "3px"))
+        self._props["height"] = value
+        self.indicator_dom_node.style.setProperty("height", _css_length(value or 3))
 
     @property
     def track_colour(self):
--- a/client_code/ProgressBar/Determinate/__init__.py
+++ b/client_code/ProgressBar/Determinate/__init__.py
@@ -19,8 +19,10 @@ _html_injector.css(ProgressBar.css)
 class Determinate(DeterminateTemplate):
     def __init__(self, **properties):
         self.indicator_dom_node = get_dom_node(self.indicator_panel)
+        self.dom_node = get_dom_node(self)
         self.role = "ae-progress-track"
         self.indicator_panel.role = "ae-progress-indicator"
+
         self._props = properties
         self.init_components(**properties)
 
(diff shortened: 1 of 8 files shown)


client_code/ProgressBar/Determinate/__init__.py
```diff
@@ -33,6 +35,24 @@ def progress(self, value):
...
+        self.indicator_dom_node.style.setProperty("height", value)
+
+    @property
+    def width(self):
+        return self._width
+
+    @width.setter
+    def width(self, value):
+        self._width = value
+        self.dom_node.style.setProperty("width", value)
```
s-cork, line 54: There's a utility function for this so we can support number as well as strings. 

It's in _component_helpers, _css_length.

Duncanr-glitch: Perfect, that makes sense. I did modify that utility function so that strings with the unit already applied can be used as well.

client_code/utils/_component_helpers.py
```diff
@@ -207,6 +208,10 @@ def walk(component_or_components):
 
 
 def _css_length(v):
+    if isinstance(v, str) and re.match(
+        r"\b\d+(\.\d+)?\s*(cm|mm|in|pt|pc|px|em|rem|vw|vh|vmin|vmax|%)\b", v
+    ):
+        return v
```
s-cork, line 214: We dont need this. The try will fail in the next step and will return the original value.

Duncanr-glitch: Yeah, I misread that block, just reverted that part.

client_code/ProgressBar/__init__.py
```diff
@@ -8,24 +8,14 @@
...
 }
 
-.anvil-role-ae-progress-indicator {
-    top: 0 !important;
-}
-
-.anvil-role-ae-progress-track > .holder, .anvil-role-ae-progress-indicator > .holder {
-    display: block !important;
-}
-
```
s-cork, line 26: Do we know why we're removing this? Doesn't seem related to the PR?

client_code/ProgressBar/Indeterminate/__init__.py
```diff
@@ -18,12 +18,33 @@
...
 
+    @property
+    def height(self):
+        return self._height
+
+    @height.setter
+    def height(self, value):
+        value = _css_length(value)
+        self._height = value
+        self.indicator_dom_node.style.setProperty("height", value)
```
s-cork, line 36: Better to convert to css length after we assign it to the private variable. That way the value you set is the value you get. This will also mean the designer property editors don't mysteriously change which I think they would currently. If you set the value to 10 then the designer might display 10px.

client_code/ProgressBar/__init__.py
Duncanr-glitch: When I just tried setting height properties with it left it and it was making the code not work or some awkward height settings.

client_code/ProgressBar/__init__.py
```diff
@@ -8,24 +8,14 @@
 
 css = """ .anvil-role-ae-progress-track, .anvil-role-ae-progress-indicator {
     display: block;
-    height: 3px;
```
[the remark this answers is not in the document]
Duncanr-glitch: I think it makes more sense to have the default in the designer, then its easier to track because it's always getting set in code imho.

client_code/ProgressBar/Indeterminate/__init__.py
Duncanr-glitch: Fair enough, I'll do that in the next push.

client_code/ProgressBar/__init__.py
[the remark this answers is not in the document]
Duncanr-glitch: I can add some null-value handling that would then default to 3px as well :)

client_code/ProgressBar/Determinate/__init__.py
```diff
@@ -33,6 +35,24 @@ def progress(self, value):
...
+        self.indicator_dom_node.style.setProperty("height", _css_length(value or "3px"))
+
+    @property
+    def width(self):
+        return self._width
+
+    @width.setter
+    def width(self, value):
+        self._width = value
+        self.dom_node.style.setProperty("width", _css_length(value or "3px"))
```
s-cork, line 54: Let's fix the default widths 😊

Duncanr-glitch: In what way do you mean? moving the or statement outside the function call, adding a default constant variable or readding it to the css?

Duncanr-glitch: Oh right, that one should be 100%, my bad. I'll do that tomorrow, already shut down the computer for the night, haha.

Duncanr-glitch: Just di that in the latest push, lmk if there's anything else :)

s-cork (approved):

Issue #380, opened 2022-11-02
Support foreground (and background) property for MultiSelectDropdown


luxint:
Hi,

I noticed that i can set the background property (in code) of  MultiSelectDropdown and then the background is set of the cell in the datagrid of which it is part of, however this doesn't work for the foreground property and when clicked both background and foreground color are white&black. Would be great if we could set these properties like the standard dropdown component.

ArneBenedikt:
I second to that. To add a color "role" or something similar to the MultiSelectDropdown would be awesome!

PR #590, opened 2024-12-23
Multiselect-dropdown-foreground-background

Duncanr-glitch:
Implements Feature #380 for background and foreground of the multiselect dropdown.

The same foreground and background are applied to the exterior and interior of the dropdown, as well as the checked icon.

The popover can also now have a background since this is how the multiselectdropdown is implemented

### DIFF
--- a/client_code/MultiSelectDropDown/Option/__init__.py
+++ b/client_code/MultiSelectDropDown/Option/__init__.py
@@ -6,6 +6,7 @@
 # This software is published at https://github.com/anvilistas/anvil-extras
 from anvil import HtmlPanel as _HtmlPanel
 from anvil.js import get_dom_node
+from anvil.js.window import jQuery as _S
 
 from ...utils._component_helpers import _add_roles, _remove_roles
 from ._anvil_designer import OptionTemplate
@@ -81,6 +82,15 @@ class Option(OptionTemplate):
             self._dom_node.dataset["disabled"] = ""
         else:
             del self._dom_node.dataset["disabled"]
+
+    @property
+    def foreground(self):
+        return self._foreground
+
+    @foreground.setter
+    def foreground(self, value):
+        self._foreground = value
+        self.label.foreground = self.icon_checked.foreground = value
 
     def focus(self):
         self._dom_node.focus()
--- a/client_code/MultiSelectDropDown/Option/__init__.py
+++ b/client_code/MultiSelectDropDown/Option/__init__.py
@@ -6,7 +6,6 @@
 # This software is published at https://github.com/anvilistas/anvil-extras
 from anvil import HtmlPanel as _HtmlPanel
 from anvil.js import get_dom_node
-from anvil.js.window import jQuery as _S
 
 from ...utils._component_helpers import _add_roles, _remove_roles
 from ._anvil_designer import OptionTemplate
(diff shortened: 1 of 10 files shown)


client_code/MultiSelectDropDown/Option/__init__.py
```diff
@@ -82,6 +83,15 @@ def disabled(self, val):
...
 
+    @property
+    def foreground(self):
+        return self._foreground
+
+    @foreground.setter
+    def foreground(self, value):
+        self._foreground = value
+        self.label.foreground = self.icon_checked.foreground = value
+
```
s-cork, line 94: Can we follow the code style of the component and keep props in the `_props` dict. 

Same for other components in this PR.

client_code/popover.py
```diff
@@ -330,6 +334,9 @@ def init_popover(self, element):
         element.setAttribute("ae-popover", "")
         element.setAttribute("ae-popover-id", self.id)
 
+        if self.background:
+            _S(element).css("background-color", self.background)
```
s-cork, line 338: Let's not use jquery when vanilla JavaScript will do just fine. Likewise for other uses of jquery

client_code/MultiSelectDropDown/__init__.py
```diff
@@ -228,6 +224,45 @@ def format_selected_text(self, count, total):
...
+        )
+
+    @property
+    def foreground(self):
+        return self._foreground
+
+    @foreground.setter
+    def foreground(self, value):
+        self._foreground = value
+        if value and self._props_initialized:
```
s-cork, line 259: It doesn't seem like it's possible to remove background/foreground

client_code/MultiSelectDropDown/__init__.py
```diff
@@ -228,6 +224,45 @@ def format_selected_text(self, count, total):
...
+            self._select_btn,
+            self._dd,
+            placement="bottom-start",
+            arrow=False,
+            delay=0,
+            animation=False,
+            trigger="manual",
+            max_width="fit-content",
+            background=value,
+        )
```
s-cork, line 250: This feels a bit of a strange place to do popover creation.

s-cork (commented):
Just some initial thoughts from looking at the code. Won't be able to properly try out these PRs till the New Year

client_code/MultiSelectDropDown/__init__.py
Duncanr-glitch: I fixed this by changing the popover so that the popover function returns the Popover object and adding background changing to the fui module's auto_update function

client_code/MultiSelectDropDown/__init__.py
Duncanr-glitch: did that by just removing the conditional checks (for the most part)

client_code/popover.py
Duncanr-glitch: done

client_code/MultiSelectDropDown/Option/__init__.py
Duncanr-glitch: done

client_code/popover.py
```diff
@@ -287,6 +289,41 @@ def __init__(
...
+    def background(self, value):
+        self._background = value
+        fui.auto_update(
+            _get_popper_element(self.popper),
+            self.dom_popover,
+            placement=self.placement,
+            arrow=self.dom_arrow if self.arrow else None,
+            foreground=self.foreground,
+            background=value,
+        )
```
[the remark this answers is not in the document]
Duncanr-glitch: That's true, I hadn't considered that, but just simplified it with that :)

s-cork (approved):

PR #599, opened 2025-03-26
Better get color

s-cork:
This PR allows theme colors to be dynamic

E.g. if you set the tab background color to theme:Primary
then when you change `theme_colors["Primary"]` the tab background will just change as expected

By using a TextBox as a dummy node, we let anvil tell us what color is being used under the hood
If it's a theme color then we get a `var(--primary-obsucre-reference-v942)` 
Which we can now use 🪄

### DIFF
--- a/.github/workflows/check_codebase.yml
+++ b/.github/workflows/check_codebase.yml
@@ -9,33 +9,31 @@ jobs:
     name: Check code base
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


meatballs:
sneaky!

meatballs:
gh actions don't like python 3.7 any more. It probably needs ubuntu-latest changing to something specific and older.

meatballs (approved):

PR #600, opened 2025-03-27
Set hook

s-cork:
this PR is an idea for a set hook

for custom components we already have descriptors on the form template
but if we want to do something when we set a property we end up recreating descriptors
what if we could just tap into the existing descriptors on the template

this PR shows that idea

```python
#Before

class C(CTemplate):
    @property
    def text(self):
        return getattr(self, "_text", None)
        
    @text.setter
    def text(self, value):
        self._text = value
        self._some_node.textContent = value

# After

class C(CTemplate):
    @set_hook
    def on_set_text(self):
        self._some_node.textContent = self.text
```

Is it worth it?
Maybe


meatballs:
I like how much code that removes!

PR #607, opened 2025-08-01
Sets Multipicker CSS ae-ms-options ul overflow to auto

danbolinson-bt:
This adopts the auto-scroll property in the multi-select dropdown suggested in this discussion item:

https://github.com/anvilistas/anvil-extras/discussions/606


s-cork:
looks good, can we get a change log entry

add it at the top under `# Unreleased`

PR #615, opened 2025-09-09
implement autocomplete with pluggable_ui textbox

s-cork:
Considerations: this is not documented api

Since m3 overrides the pluggable_ui["anvil.TextBox"]
It means that anyone using an autocomplete component in an m3 app will get an m3 textbox instead of an anvil textbox,
which is probably what they want, but might be unexpected
especially if they've added some css

Might be considered breaking, but I think it's probably an enhancement??

Also adds virtualization, since we added it for the other list rendering component, why not

addresses discussion #603

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# Unreleased
+
+## Enhancements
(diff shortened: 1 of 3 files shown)


s-cork:

- #615 ◀

PR #616, opened 2025-09-10
Fix: wrap view transition in try except to mitigate against random errors

s-cork:
not sure this is the right fix, but wrapping it in a try except might work

trying to avoid the random error about invalid state
see:
https://anvil.works/forum/t/anyone-seen-invalidstateerror-view-transition-was-skipped-because-document-visibility-state-is-hidden-with-anvil-router/24990/6


s-cork:

- #616 ◀

s-cork:
small change and bug fix, merging

PR #627, opened 2025-09-26
Fix: chrome tablet focus issue

### LINKED ISSUE #626: Issue with Anvil Extras Multi-Select Dropdown on Chrome for Tablet
I'm encountering an issue with the latest version of Anvil Extras (3.4.1) when using the multi-select dropdown in Chrome on a tablet (latest version of Chrome). Tapping the dropdown causes the on-screen keyboard to briefly appear and then immediately close, approximately 9 out of 10 times. I haven’t been able to determine what triggers the keyboard to remain open in the rare case when it does.

This behavior makes it nearly impossible to use the keyboard for searching, forcing users to scroll through the list instead.

Everything works as expected in desktop versions of Chrome, Safari, Firefox, etc., and I did not experience this issue with earlier versions of Anvil Extras. That said, the last time I tested the app on an Android tablet using Chrome was around six months ago.

Thanks and best regards,
Arne

s-cork:
a quick fix for preventing the search filter box from losing focus immediately after it receives focus

fixes #626

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Bug Fixes
+- multiselect - fix bug on chrome tablet where focus was lost from search box
+  https://github.com/anvilistas/anvil-extras/pull/627
+
 # v3.4.1
 
 ## Bug Fixes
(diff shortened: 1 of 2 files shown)


s-cork:

- #627 ◀

PR #629, opened 2025-10-08
Routing: fix circular import issues with lazy_route

s-cork:
If a lazy route results in a module that takes a while to load
then a trigger happy user can cause this lazy route to be called multiple times
this results in false positive circular import errors occuring
(whilst a module is being imported, the same module gets imported)

This PR prevents inflight calls to lazy routes by using the existing non_blocking machinery
In this PR we can only ever have one inflight call to the same lazy route


s-cork:

- #629 ◀

PR #633, opened 2025-11-19
Feat: opt out of cdn loading

s-cork:
**Refactor asset loading with CDN fallback and local asset support**

- Centralized asset loading in `_cdn_loader.py` with semantic configuration
- Supports ESM, CJS, and window-global JS types
- CDN with automatic fallback to local assets
- Separated CDN/local configs to handle different module types (e.g., CDN ESM → local UMD)
- Bundled all assets locally in `/theme/assets/anvil-extras/cdn`
- Migrated from Skypack to jsDelivr
- Updated Quill, Floating UI, noUiSlider, localforage, and TanStack Virtual to use the new loader
- Added documentation for CDN configuration

Users can configure CDN usage via Settings → Dependencies → anvil-extras → Client → cdn (true/false).

Testing:
- local testing

Small Change:
- we used to load the css if the window object already existed
- now we don't load the css if the window object exists
- so someone who put quill js in their native libraries but didn't put the css, will have an unstyled quill component
- i'm ok with that though, it's an easy fix for a user of anvil-extras


s-cork:
it's new, not sure if it's even documented

but if you define your client config in the yaml
then in the IDE you will see config options when clicking edit on a dependency

it's fairly handy for this kinda thing

meatballs (approved):

PR #644, opened 2026-01-11
Allow has_permisssion() and check_permissions() to take a keyword arg user

kaidenwang0309-hub:
I added a function in authorisation.py that checks if multiple users have permission to call a function (returns a dict). I'm not sure if I should make the function semi-private. Please let me know your thoughts.

I think this function could be useful for admin panels, workflow apps to see who can access certain information, etc..

### DIFF
--- a/server_code/authorisation.py
+++ b/server_code/authorisation.py
@@ -74,8 +74,7 @@ def has_permission(permissions, user=sentinel):
 
 def check_permissions(permissions, user=sentinel):
     """Checks a users permissions, raises ValueError if user does not have permissions"""
-    if user is sentinel:
-        user = anvil.users.get_user()
+    user = anvil.users.get_user() if user is sentinel else None
 
     if has_permission(permissions, user=user):
         return


kaidenwang0309-hub:
I've made the changes, but it doesn't appear on my side of the PR. It does show when I open my branch, so if you are facing the same issue, the changes should show up in my branch. I believe the PR will still include the new changes, but I wanted to let you know.

s-cork:
I don't really see the use case for the `user_permissions_dict` 

Doesn't that just obfuscate whether all the users have permission or not?

Can you provide an example of how you'd be using this function?

s-cork:
I'm concerned about the API surface area here. If `each_user_has_permission` is intended to filter users, the name is a bit misleading—I’d expect a boolean return based on the name.

If the goal is filtering, a list comprehension is already very idiomatic:

```python
users_who_can_comment = [
    u for user in all_users
        if has_permission(permissions, user=u)
]

# or with filter

can_comment = partial(has_permission, comment_permissions)
users_who_can_comment = filter(can_comment, all_users)

```

I'd consider adding the same sentinel approach to `check_user` so the api is consistent across both functions.

kaidenwang0309-hub:
I refactored the check_permissions() logic to include the sentinel approach.

server_code/authorisation.py
```diff
@@ -46,9 +48,11 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-def has_permission(permissions):
+def has_permission(permissions, user=sentinel):
     """Returns True/False depending on whether a user has permission or not"""
-    user = anvil.users.get_user()
+    if user == sentinel:
```
s-cork: Let's use `user is sentinel` rather than `==`.

server_code/authorisation.py
```diff
@@ -69,12 +73,14 @@ def has_permission(permissions):
     return required_permissions.issubset(user_permissions)
 
 
-def check_permissions(permissions):
+def check_permissions(permissions, user=sentinel):
     """Checks a users permissions, raises ValueError if user does not have permissions"""
-    if has_permission(permissions):
+    if user == sentinel:
```
s-cork: Likewise

s-cork (commented):
One suggestion and then... 

To get this merged we'd need to update the docs in authorisation.rst

An LLM can probably do that for you. 

And also update the Changelog.md


Add a new heading at the top of that file 

```markdown
# Unreleased

## features
- <short description>
  https://github.com/anvilistas/anvil-extras/pull/644
```

kaidenwang0309-hub:
I've made the changes you suggested, updated the Changelog.md, and updated authorisation.rst.

server_code/authorisation.py
```diff
@@ -46,9 +48,11 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-def has_permission(permissions):
+def has_permission(permissions, user=sentinel):
     """Returns True/False depending on whether a user has permission or not"""
-    user = anvil.users.get_user()
+    if user is sentinel:
+        user = anvil.users.get_user()
```
meatballs:
```suggestion
    user = anvil.users.get_user() if user is sentinel else user
```
slightly more declarative

server_code/authorisation.py
```diff
@@ -69,12 +73,14 @@ def has_permission(permissions):
     return required_permissions.issubset(user_permissions)
 
 
-def check_permissions(permissions):
+def check_permissions(permissions, user=sentinel):
     """Checks a users permissions, raises ValueError if user does not have permissions"""
-    if has_permission(permissions):
+    if user is sentinel:
+        user = anvil.users.get_user()
```
meatballs:
```suggestion
    user = anvil.users.get_user() if user is sentinel else None
```

meatballs (commented):
Minor suggestions from me but otherwise looks good!

PR #645, opened 2026-01-13
Simplified code across multiple modules

kaidenwang0309-hub:
I am not too familiar with Anvil, so my changes may affect what the code actually does, so please do let me know about any fixes I may need to make.


meatballs:
We're all volunteers here. We aren't about to spend time reviewing something where you can't explain its effect.

